PR #8, opened 2022-02-22
Provide compatible coroutine.h between compiler and std library

ChuanqiXu9:
The current implementation for coroutine.h would be problematic if we
use the combination of clang and libstdc++ (version > 10). Provide
solution for it.


qicosmos (approved):
LGTM

PR #9, opened 2022-02-22
Basic CI Support

ChuanqiXu9:
## Why

Basic CI support for each PR.


qicosmos:
I think we also need fix a clang-format version and clang-format checking.

ChuanqiXu9:
> I think we also need to fix a clang-format version and clang-format checking.

This is listed in another issue.

qicosmos (approved):
LGTM

Issue #16, opened 2022-02-22
assign a Promise to another existing Promise cause ref count leak


RainMark:
### Search before asking

- [ ] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

### What happened + What you expected to happen

The Promise copy assignment implement missing invoke detachPromise()

### Reproduction way

We construct a unit test, expect to got "Promise is broken" exception, but it's failed
```
    auto promise1 = std::make_unique<Promise<int>>();
    auto promise2 = std::make_unique<Promise<int>>();
    promise2->setValue(0);
    auto future = promise1->getFuture();
    *promise1 = *promise2;
    promise1.reset();
    ASSERT_THROW(future.value(), std::runtime_error);
```

### Anything else

I try to delete the Promise's copy assignment operator, but failed.
Looks like Executor::schedule(Func func) broken the promise's move construct if promise object captured by lambda.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #20, opened 2022-02-22
fix Promise copy assign operator

RainMark:

## Why

fix for #16

### DIFF
--- a/async_simple/Promise.h
+++ b/async_simple/Promise.h
@@ -35,7 +35,7 @@ template <typename T>
 class Promise {
 public:
     Promise() : _sharedState(new FutureState<T>()), _hasFuture(false) {
-        _sharedState->attachPromise();
+        _sharedState->attachOne();
     }
     ~Promise() {
         if (_sharedState) {
@@ -43,18 +43,10 @@ public:
         }
     }
 
-    Promise(const Promise& other) {
-        _sharedState = other._sharedState;
-        _hasFuture = other._hasFuture;
-        _sharedState->attachPromise();
-    }
-    Promise& operator=(const Promise& other) {
-        this->~Promise();
-        _sharedState = other._sharedState;
-        _hasFuture = other._hasFuture;
-        _sharedState->attachPromise();
-        return *this;
-    }
+    // std::function require captured objects be copyable
+    // use MoveWrapper to workaround
+    Promise(const Promise& other) = delete;
+    Promise& operator=(const Promise& other) = delete;
 
     Promise(Promise<T>&& other)
         : _sharedState(other._sharedState), _hasFuture(other._hasFuture) {
--- a/async_simple/Promise.h
+++ b/async_simple/Promise.h
@@ -49,6 +49,7 @@ public:
         _sharedState->attachPromise();
     }
     Promise& operator=(const Promise& other) {
+        this->~Promise();
         _sharedState = other._sharedState;
         _hasFuture = other._hasFuture;
         _sharedState->attachPromise();
@@ -103,9 +104,8 @@ public:
     }
 
 private:
-private:
-    FutureState<T>* _sharedState;
-    bool _hasFuture;
+    FutureState<T>* _sharedState = nullptr;
+    bool _hasFuture = false;
 };
 
 }  // namespace async_simple
--- a/async_simple/Promise.h
+++ b/async_simple/Promise.h
@@ -35,7 +35,7 @@ template <typename T>
 class Promise {
 public:
     Promise() : _sharedState(new FutureState<T>()), _hasFuture(false) {
-        _sharedState->attachOne();
+        _sharedState->attachPromise();
     }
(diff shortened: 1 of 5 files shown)


async_simple/Promise.h
```diff
@@ -49,6 +49,9 @@ class Promise {
         _sharedState->attachPromise();
     }
     Promise& operator=(const Promise& other) {
+        if (_sharedState) {
```
ChuanqiXu9, line 52: I prefer `this->~Promise();`. This is more stable if we add other logic to `~Promise()` then.

RainMark: 👍🏻，updated

foreverhy: Remove copy constructor is also acceptable, I believe passing a promise to another attached promise is not a good choice.
The only reason is std::function requires captured objects should be copyable, but we can use MoveWrapper here.

ChuanqiXu9: Agreed. This is what Seastar do now: https://github.com/scylladb/seastar/blob/ea6a6820ede9995cbfc6037e430b8d733a088c8f/include/seastar/core/future.hh#L832-L833.

RainMark: ok，I try to remove it again.

ChuanqiXu9: @RainMark sorry, I took an oversight. Seastar didn't delete the copy assignment. It just defines it out-of-line. https://github.com/scylladb/seastar/blob/ea6a6820ede9995cbfc6037e430b8d733a088c8f/include/seastar/core/future.hh#L973-L979

ChuanqiXu9 (approved):
LGTM.

async_simple/Promise.h
ChuanqiXu9: Copy assignment is used in `forward_to`: https://github.com/scylladb/seastar/blob/ea6a6820ede9995cbfc6037e430b8d733a088c8f/include/seastar/core/future.hh#L1721

RainMark: @ChuanqiXu9 your link pointed to move assignment?

RainMark: looks like seastar use noncopyable function to workaround? https://github.com/scylladb/seastar/blob/ea6a6820ede9995cbfc6037e430b8d733a088c8f/include/seastar/util/noncopyable_function.hh#L102

ChuanqiXu9: Oh, my bad. Thanks for pointing me.

ChuanqiXu9: > looks like seastar use noncopyable function to workaround? https://github.com/scylladb/seastar/blob/ea6a6820ede9995cbfc6037e430b8d733a088c8f/include/seastar/util/noncopyable_function.hh#L102

From its comment, it looks like our MoveWrapper, is it?

RainMark: @ChuanqiXu9 em, not same, them not use std::function, so no need to wrap the promise

such as the timer's callback: https://github.com/scylladb/seastar/blob/ea6a6820ede9995cbfc6037e430b8d733a088c8f/include/seastar/core/timer.hh#L84

but we already used the std::function in base class interface, so we keep the current usage

async_simple/uthread/test/UthreadTest.cpp
```diff
@@ -111,9 +111,9 @@ TEST_F(UthreadTest, testSwitch) {
         Promise<int> p;
         auto f = p.getFuture().via(&_executor);
         delayedTask(
-            [p = std::move(p)]() mutable {
+            [wrapper = MoveWrapper(std::move(p))]() mutable {
```
ChuanqiXu9, line 114: I feel such change would make it harder to use. I am sure seastar has copy constructor/assignment for promise before. I am wondering if it is better to fix the copy constructor/assignment **now**. And delete it later when we introduce `noncopyable_function `.

RainMark: we can create another issue to remove MoveWrapper, maybe use` noncopyable_function` or the other solution

ChuanqiXu9: FutureState.h is internal too. It doesn't matter if we write some not direct code. But now user need to write `MoveWrapper` in some cases, it is not straightforward. I prefer to fix the copy constructor/assignment in this PR and create an issue to introduce `nocopyable_function` and delete copy constructor/assignment.

RainMark: @foreverhy how do you think?

qicosmos: Actually i tried to remove MoveWrapper before, replace it with move capture, but compile failed, maybe need to do more changes.
I wish we can use move capture instead of MoveWrapper.

RainMark: emm, I agree to fix the copy assignment operator firstly

PR #23, opened 2022-02-23
[Executor interface]mod interface

qicosmos:
## Why

Make custom executor definition more easier.

## What is changing

Modify some pure virtual functions to non pure virtual functions, custom executor only need to override only one function, we have to override 4 functions before.

### DIFF
--- a/async_simple/Executor.h
+++ b/async_simple/Executor.h
@@ -88,8 +88,8 @@ public:
     // should guarantee that the func would be executed.
     virtual bool schedule(Func func) = 0;
     // Return true if caller runs in the executor.
-    virtual bool currentThreadInExecutor() const = 0;
-    virtual ExecutorStat stat() const = 0;
+    virtual bool currentThreadInExecutor() const { return false; }
+    virtual ExecutorStat stat() const { return async_simple::ExecutorStat(); }
 
     // checkout() return current "Context", which defined by executor
     // implementation, then checkin(func, "Context") should schdule func to the
@@ -119,7 +119,7 @@ public:
 
     // IOExecutor accepts IO read/write requests.
     // Return nullptr if the exeuctor doesn't offer an IOExecutor.
-    virtual IOExecutor *getIOExecutor() = 0;
+    virtual IOExecutor *getIOExecutor() { return nullptr; }
 
     // This method will block current thread until func complete.
     // Return false if it fails to schedule func. Return true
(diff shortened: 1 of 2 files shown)


async_simple/Executor.h
```diff
@@ -88,8 +88,8 @@ class Executor {
     // should guarantee that the func would be executed.
     virtual bool schedule(Func func) = 0;
     // Return true if caller runs in the executor.
-    virtual bool currentThreadInExecutor() const = 0;
-    virtual ExecutorStat stat() const = 0;
+    virtual bool currentThreadInExecutor() const { return false; }
```
RainMark, line 91: Future/Promise components used this interface to check current context, If user used Future/Promise may got unexpected error with this default implement, how about to throw the exception to tell user need to implement it?

qicosmos: Good, could you give advice about throwing what kind of exception.

RainMark: std::logic_error ?

qicosmos: OK

async_simple/Executor.h
```diff
@@ -119,7 +121,7 @@ class Executor {
 
     // IOExecutor accepts IO read/write requests.
     // Return nullptr if the exeuctor doesn't offer an IOExecutor.
-    virtual IOExecutor *getIOExecutor() = 0;
+    virtual IOExecutor *getIOExecutor() { return nullptr; }
```
ChuanqiXu9, line 124: I think this one should throw error too.

async_simple/Executor.h
```diff
@@ -88,8 +88,10 @@ class Executor {
     // should guarantee that the func would be executed.
     virtual bool schedule(Func func) = 0;
     // Return true if caller runs in the executor.
-    virtual bool currentThreadInExecutor() const = 0;
-    virtual ExecutorStat stat() const = 0;
+    virtual bool currentThreadInExecutor() const {
+        throw std::logic_error("Not implemented");
+    }
+    virtual ExecutorStat stat() const { return async_simple::ExecutorStat(); }
```
ChuanqiXu9, line 94: ditto

async_simple/Executor.h
qicosmos: ok

RainMark (approved):

PR #25, opened 2022-02-25
add support for gcc

### LINKED ISSUE #11: Support GCC
### Search before asking

async_simple only support clang now, we also need to support gcc.

### Anything else
#25 

foreverhy:
## Why

Close #11 

add gcc support.

### DIFF
--- a/async_simple/coro/Collect.h
+++ b/async_simple/coro/Collect.h
@@ -77,7 +77,8 @@ struct CollectAnyAwaiter {
         : _input(std::move(other._input)), _result(std::move(other._result)) {}
 
     bool await_ready() const noexcept {
-        return _input.empty() || (_result && _result->_idx != -1);
+        return _input.empty() ||
+               (_result && _result->_idx != static_cast<size_t>(-1));
     }
 
     void await_suspend(STD_CORO::coroutine_handle<> continuation) {
@@ -91,10 +92,11 @@ struct CollectAnyAwaiter {
         std::vector<LazyType, InAlloc> input(std::move(_input));
         auto result = std::make_shared<ResultType>();
         auto event = std::make_shared<detail::CountEvent>(input.size());
-        size_t i = 0;
 
         _result = result;
-        for (i = 0; i < input.size() && (result->_idx == -1); ++i) {
+        for (size_t i = 0;
+             i < input.size() && (result->_idx == static_cast<size_t>(-1));
+             ++i) {
             if (!input[i]._coro.promise()._executor) {
                 input[i]._coro.promise()._executor = executor;
             }
--- a/async_simple/coro/Task.h
+++ b/async_simple/coro/Task.h
@@ -139,10 +139,8 @@ struct TaskPromiseBase {
     }
     void unhandled_exception() const noexcept {}
 
-    void* operator new(size_t sz) noexcept { return ::operator new(sz); }
-    void operator delete(void* p, size_t sz) noexcept {
-        return ::operator delete(p);
-    }
+    void* operator new(size_t sz) { return ::operator new(sz); }
+    void operator delete(void* p, size_t sz) { return ::operator delete(p); }
 };
 
 template <typename T>
--- a/async_simple/coro/test/LazyTest.cpp
+++ b/async_simple/coro/test/LazyTest.cpp
@@ -418,7 +418,7 @@ TEST_F(LazyTest, testCollectAll) {
 }
 
 TEST_F(LazyTest, testCollectAllBatched) {
-    size_t task_num = 5000;
+    int task_num = 5000;
     long total = 0;
     for (auto i = 0; i < task_num; i++) {
         total += i;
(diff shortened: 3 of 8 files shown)


async_simple/coro/Task.h
```diff
@@ -139,8 +139,8 @@ struct TaskPromiseBase {
     }
     void unhandled_exception() const noexcept {}
 
-    void* operator new(size_t sz) noexcept { return ::operator new(sz); }
-    void operator delete(void* p, size_t sz) noexcept {
+    void* operator new(size_t sz) { return ::operator new(sz); }
+    void operator delete(void* p, size_t sz) {
```
ChuanqiXu9, line 143: Why do we need to remove the `noexcept` specifier?

async_simple/coro/test/LazyTest.cpp
```diff
@@ -420,7 +420,7 @@ TEST_F(LazyTest, testCollectAll) {
 TEST_F(LazyTest, testCollectAllBatched) {
     size_t task_num = 5000;
     long total = 0;
-    for (auto i = 0; i < task_num; i++) {
+    for (int i = 0; i < static_cast<int>(task_num); i++) {
```
ChuanqiXu9, line 423: Is this change necessary?

ChuanqiXu9:
It looks like this patch need to be formatted by `git-clang-format HEAD^`.

async_simple/coro/Collect.h
```diff
@@ -94,7 +94,7 @@ struct CollectAnyAwaiter {
         size_t i = 0;
```
qicosmos, line 94: ```size_t i = 0``` in for-loop would be better?

async_simple/coro/test/LazyTest.cpp
```diff
@@ -420,7 +420,7 @@ TEST_F(LazyTest, testCollectAll) {
 TEST_F(LazyTest, testCollectAllBatched) {
     size_t task_num = 5000;
     long total = 0;
```
qicosmos, line 422: ```long total``` --> ```size_t total``` may be better.
Maybe unified all int to size_t can get ride of static_cast.

async_simple/coro/Task.h
foreverhy: Not compile in gcc, Task::promise_type do not have a get_return_object_on_allocation_failure().

async_simple/coro/test/LazyTest.cpp
foreverhy: make task_num from `size_t` to `int`.

async_simple/coro/Task.h
ChuanqiXu9: It would be better to add `get_return_object_on_allocation_failure()` methods. It could return a Lazy with an exception shows it fails on allocation failure.

async_simple/coro/test/LazyTest.cpp
ChuanqiXu9: ping?

async_simple/coro/Task.h
foreverhy: Task do not have an exception_ptr member, which is different from Lazy.

async_simple/coro/test/LazyTest.cpp
foreverhy: now task_num from `size_t` to  `int`, static_cast is unnecessary.

async_simple/coro/Task.h
ChuanqiXu9: How about adding `exception_ptr` for Task? Would it be too hard?

async_simple/coro/test/LazyTest.cpp
```diff
@@ -448,7 +448,7 @@ TEST_F(LazyTest, testCollectAllBatched) {
         EXPECT_EQ(task_num, out.size());
         co_await CurrentExecutor();
         int sum = 0;
-        for (auto i = 0; i < out.size(); i++) {
+        for (size_t i = 0; i < out.size(); i++) {
```
ChuanqiXu9, line 451: Is this necessary? Does GCC warn if we use `auto` here?

async_simple/coro/Task.h
foreverhy: `Task` was created for in-memory applications, maybe we could remove it since nobody is using it.

async_simple/coro/test/LazyTest.cpp
foreverhy: `auto` is deduced to `int` here.

qicosmos (approved):
LGTM

async_simple/coro/Task.h
ChuanqiXu9: If nobody uses it, I agree we should remove it.

async_simple/coro/test/LazyTest.cpp
ChuanqiXu9: Got it. We'd better use range-based for here. But we could do it in next patches.

async_simple/coro/Task.h
foreverhy: Will fix in #29

ChuanqiXu9: OK

Issue #33, opened 2022-03-03
Remove util/Condition.h


RainMark:
### Search before asking

- [ ] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

### What happened + What you expected to happen

util/Condition.h like c++ std::condition_variable, we better use std::condition_variable to keep cross platform

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

qicosmos:
Done.

4kangjc:
现在的语义是信号量？

4kangjc:
> 现在的语义是信号量？

https://en.cppreference.com/w/cpp/thread/counting_semaphore

ChuanqiXu9:
> > 现在的语义是信号量？
> 
> https://en.cppreference.com/w/cpp/thread/counting_semaphore

是的，现在是在 std::condition_variable 上简单包了一层

4kangjc:
> > > 现在的语义是信号量？
> > 
> > 
> > https://en.cppreference.com/w/cpp/thread/counting_semaphore
> 
> 是的，现在是在 std::condition_variable 上简单包了一层

```cpp
class Condition : public std::binary_semaphore {
public:
    explicit Condition(ptrdiff_t num = 0) : std::binary_semaphore(num) {}
};
```
可惜某些编译器还不太能支持binary_semaphore

ChuanqiXu9:
> https://en.cppreference.com/w/cpp/thread/counting_semaphore

这样，不过我觉得这个倒是影响不大；不过感兴趣的话可以新开个 issue 来 track

PR #36, opened 2022-03-03
[Cross platform]use std::condition_variable

qicosmos:
## Why

We'd better avoid use specific os api to implement a Condition, we can use std::condition_variable to implement this to support cross platform.

Actually c++20 has already provided [semaphore](https://en.cppreference.com/w/cpp/thread/counting_semaphore) , but clang has not supported it yet, however we could keep the same api with c++20 semaphore, so the implementation of Condition.h and API changes. We can remove Condition in future when the compilers are ready for semaphore.

for #33 

## What is changing

Remove futes dependency, change the APIs according to c++20 [semaphore](https://en.cppreference.com/w/cpp/thread/counting_semaphore).

### DIFF
--- a/async_simple/Executor.h
+++ b/async_simple/Executor.h
@@ -134,12 +134,12 @@ public:
         util::Condition cond;
         auto ret = schedule([f = std::move(func), &cond]() {
             f();
-            cond.set();
+            cond.release();
         });
         if (!ret) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


ChuanqiXu9 (approved):
LGTM.

RainMark (approved):

Issue #41, opened 2022-03-06
CountChar build failed in gcc 11.2.1


vbirds:
gcc version:
gcc (GCC) 11.2.1 20210728 (Red Hat 11.2.1-1)
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE

[ 71%] Built target async_simple
[ 85%] Building CXX object demo_example/CMakeFiles/CountChar.dir/CountChar.cpp.o
In file included from /tmp/tmp.XuLZlZqHI1/async_simple/coro/LazyHelper.h:56,
                 from /tmp/tmp.XuLZlZqHI1/async_simple/coro/Lazy.h:538,
                 from /tmp/tmp.XuLZlZqHI1/demo_example/CountChar.cpp:20:
/tmp/tmp.XuLZlZqHI1/async_simple/coro/Collect.h: In member function ‘void async_simple::coro::detail::CollectAnyAwaiter<LazyType, IAlloc>::await_suspend(std::__n4861::coroutine_handle<void>)’:
/tmp/tmp.XuLZlZqHI1/async_simple/coro/Collect.h:85:13: error: invalid ‘static_cast’ from type ‘std::__n4861::coroutine_handle<void>*’ to type ‘std::__n4861::coroutine_handle<async_simple::coro::detail::LazyPromiseBase>*’
   85 |             static_cast<STD_CORO::coroutine_handle<LazyPromiseBase>*>(
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   86 |                 &continuation)
      |                 ~~~~~~~~~~~~~~
/tmp/tmp.XuLZlZqHI1/async_simple/coro/Collect.h: In member function ‘void async_simple::coro::detail::CollectAllAwaiter<LazyType, IAlloc, OAlloc, Para>::await_suspend(std::__n4861::coroutine_handle<void>)’:
/tmp/tmp.XuLZlZqHI1/async_simple/coro/Collect.h:161:13: error: invalid ‘static_cast’ from type ‘std::__n4861::coroutine_handle<void>*’ to type ‘std::__n4861::coroutine_handle<async_simple::coro::detail::LazyPromiseBase>*’
[...]

PR #42, opened 2022-03-07
Conform C++20 <coroutine>

ChuanqiXu9:
## Why

The previous implementation of `coroutine.h" is based on coroutine TS. And it is slightly different from C++20 <coroutine>.  This patch makes the implementation of "coroutine.h" conforming with C++20 <coroutine>.

This would solve issue https://github.com/alibaba/async_simple/issues/41 too.


qicosmos (approved):
LGTM

PR #43, opened 2022-03-07
Add Lazy benchmarks initially

ChuanqiXu9:
Currently we didn't have in-tree benchmarks and we need benchmarking CI to monitoring performance changes. This is the initial submitting which contains small benchmarks for lazy only.


qicosmos (approved):
LGTM

PR #46, opened 2022-03-08
Enable Performance CI

ChuanqiXu9:
## Why

We need monitoring every commit and PR to avoid unimagined regressions.

## What is changing

We would receive an comment if the PR or commit cause a regression greater than 50%.

## Example

https://github.com/ChuanqiXu9/async_simple/pull/5#commitcomment-68189590


qicosmos (approved):
LGTM

PR #49, opened 2022-03-08
Split Performance CI

ChuanqiXu9:
## Why

Due to the access control system, we need different actions for PR and commit. For example, we need record the performance of every commit. But we shouldn't record for every PR.


poor-circle:
Testing

PR #51, opened 2022-03-08
[coro] Add FutureAwaiter

### LINKED ISSUE #37: Avoid User Write Awaiter
### Search before asking

- [ ] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

After user used Lazy , they have to write many Awaiter to adapt the exist callback type asynchronous APIs
We can introduce a FutureAwaiter to avoid it.

eg:

```c++

template <typename T>
class FutureAwaiter {
public:
    FutureAwaiter() {
        _fut = _pr.getFuture();
    }
    void setValue(T&& value) {
        _pr.setValue(std::move(value));
    }


    bool await_ready() {
        return _fut.hasResult();
    }
    void await_suspend(impl::coroutine_handle<> h) {
        auto func = [h](Try<T>&& t) mutable {
            h.resume();
        };
        std::move(_fut).thenTry(std::move(func));
    }
    T&& await_resume() { return std::move(_fut.value()); }

private:
    Promise<T> _pr;
    Future<T> _fut;
};

// exist callback style asynchronous api
template <typename Callback>
void asyncTask(CallBack&& cb) {
    // ...
    cb(0);
}

// use example

Lazy<int> asyncTaskCoro() {
    FutureAwaiter awaiter;
    asyncTask([&awaiter](auto result) {
                  awaiter.setValue(result);
              });
    int ret = co_await awaiter;
    co_return ret;
}

```

RainMark:

## Why

Close #37 

## What is changing

add FutureAwaiter.h and Test source

## Example

```c++
Promise<T> pr;
auto fut = pr.getFuture();
// invoke pr.setValue() in callback
auto val = co_await FutureAwaiter(std::move(fut));
```

### DIFF
--- a/async_simple/coro/FutureAwaiter.h
+++ b/async_simple/coro/FutureAwaiter.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2022, Alibaba Group Holding Limited;
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef ASYNC_SIMPLE_CORO_FUTURE_AWAITER_H
+#define ASYNC_SIMPLE_CORO_FUTURE_AWAITER_H
+
+#include <async_simple/Future.h>
+#include <async_simple/experimental/coroutine.h>
+
+namespace async_simple {
+namespace coro {
+
+template <typename T>
+class FutureAwaiter {
+public:
+    explicit FutureAwaiter(Future<T>&& future) : _future(std::move(future)) {}
+    FutureAwaiter(FutureAwaiter&& rhs) : _future(std::move(rhs._future)) {}
+    FutureAwaiter(FutureAwaiter&) = delete;
+
+    bool await_ready() { return _future.hasResult(); }
+    void await_suspend(CoroHandle<> continuation) {
+        _future.setContinuation(
+            [continuation](Try<T>&& t) mutable { continuation.resume(); });
+    }
+    T await_resume() { return std::move(_future.value()); }
+
+private:
+    Future<T> _future;
+};
+
+template <typename T>
+auto operator co_await(T&& future) requires IsFuture<std::decay_t<T>>::value {
+    return FutureAwaiter(std::move(future));
+}
+
+}  // namespace coro
+}  // namespace async_simple
+
+#endif
--- a/async_simple/coro/test/FutureAwaiterTest.cpp
+++ b/async_simple/coro/test/FutureAwaiterTest.cpp
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2022, Alibaba Group Holding Limited;
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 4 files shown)


ChuanqiXu9 (commented):
How about add an operator:
```
FutureAwaiter operator co_awiat(Future)
```

So that we could co_await Future directly.

RainMark:
> How about add an operator:
> 
> ```
> FutureAwaiter operator co_awiat(Future)
> ```
> 
> So that we could co_await Future directly.

cool, I will add it

ChuanqiXu9 (commented):
How do you think about:
```
auto operator co_await(Future<T>& future) {
    return return FutureAwaiter(std::move(future));
}
```

Now user could co_await Future directly instead of `co_await std::move(Future)`. Although it breaks the style we kept before.

ChuanqiXu9:
I would close and reopen the PR to trigger the new CI.

ChuanqiXu9:
The failed Performance check is outdated. Let's omit it.

async_simple/coro/test/FutureAwaiterTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+        }).detach();
+    }
+};
+
+TEST_F(FutureAwaiterTest, testWithFuture) {
+    auto lazy1 = [&]() -> Lazy<> {
+        Promise<int> pr;
+        auto fut = pr.getFuture();
+        sum(1, 1, [pr = std::move(pr)](int val) mutable { pr.setValue(val); });
+        usleep(1000);
```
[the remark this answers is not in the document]
RainMark: ok

RainMark:
> How do you think about:
> 
> ```
> auto operator co_await(Future<T>& future) {
>     return return FutureAwaiter(std::move(future));
> }
> ```
> 
> Now user could co_await Future directly instead of `co_await std::move(Future)`. Although it breaks the style we kept before.

done

async_simple/coro/test/FutureAwaiterTest.cpp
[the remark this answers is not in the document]
qicosmos: OK

qicosmos (approved):
LGTM

async_simple/coro/FutureAwaiter.h
```diff
@@ -0,0 +1,55 @@
...
+
+private:
+    Future<T> _future;
+};
+
+template <typename T,
+          std::enable_if_t<
+              std::is_same_v<std::decay_t<T>,
+                             Future<typename std::decay_t<T>::value_type>>,
+              int> = 0>
```
ChuanqiXu9, line 47: I think we could use `IsFuture` or `std::is_reference` here.

BTW, since we are using C++20, we could use concepts even:
```
template <class T>
concept FutureTy = async_simple::IsFuture<T>::value_type;

template <FutureTy Fut>
auto operator co_await(Fut&& future) { ... }
```

The name could be changed.

async_simple/coro/FutureAwaiter.h
```diff
@@ -0,0 +1,55 @@
...
+    explicit FutureAwaiter(Future<T>&& future) : _future(std::move(future)) {}
+    FutureAwaiter(FutureAwaiter&& rhs) : _future(std::move(rhs._future)) {}
+    FutureAwaiter(FutureAwaiter&) = delete;
+
+    bool await_ready() { return _future.hasResult(); }
+    void await_suspend(CoroHandle<> continuation) {
+        _future.setContinuation(
+            [continuation](Try<T>&& t) mutable { continuation.resume(); });
+    }
+    T&& await_resume() { return std::move(_future.value()); }
```
ChuanqiXu9, line 37:
```
T await_resume() { return std::move(_future).value(); }
```

So that the compiler would do mandatory copy-elision: https://en.cppreference.com/w/cpp/language/copy_elision

RainMark: done

async_simple/coro/FutureAwaiter.h
RainMark: used requires syntax

ChuanqiXu9 (approved):
LGTM. The performance alert shouldn't be true.

PR #67, opened 2022-03-23
remove fs dependency

qicosmos:
## Why

Remove filesystem dependency from demo example.


demo_example/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ endif()
 
 add_executable(smtp_client smtp/smtp_client.cpp)
 if (ENABLE_SSL)
-    target_link_libraries(smtp_client stdc++fs pthread OpenSSL::SSL)
+    target_link_libraries(smtp_client async_simple OpenSSL::SSL)
 else()
-    target_link_libraries(smtp_client stdc++fs pthread)
+    target_link_libraries(smtp_client async_simple)
```
RainMark, line 52: have you test with uthread disabled?

RainMark: we need SRCS to build the async_simple target, Otherwise, we always thinking about whether to link it or not

ChuanqiXu9: I agree to move Condition.h into cpp. But I prefer the idea of empty source file. The empty source file would tell the user that it is intended to be so. I feel the semantics here is better.

qicosmos: I think we can use std::promise instead of Condition.h

RainMark: if use empty source file, whether all related headers both add empty source files?

ChuanqiXu9: > I think we can use std::promise instead of Condition.h

It's odd to use std::promise since we have our own promise. And promise has unnecessary wrapper too.

> if use empty source file, whether all related headers both add empty source files?

I am not sure what you mean. How could the headers add source files?

RainMark: i mean Lazy.h we add Lazy.cpp(include Lazy,h) and the other header file both  do it

RainMark: > > I think we can use std::promise instead of Condition.h
> 
> It's odd to use std::promise since we have our own promise. And promise has unnecessary wrapper too.
> 
> > if use empty source file, whether all related headers both add empty source files?
> 
> I am not sure what you mean. How could the headers add source files?

you mean completely one empty file?

ChuanqiXu9: > you mean completely one empty file?

Yeah, the semantics look more clear to me.

> i mean Lazy.h we add Lazy.cpp(include Lazy,h) and the other header file both do it

Include *.cpp in header looks more bad to me...

qicosmos: Is it possible to make coro part header only and standalone? that will be very easy to use for stackless coroutine scene.

RainMark: > > you mean completely one empty file?
> 
> Yeah, the semantics look more clear to me.
> 
> > i mean Lazy.h we add Lazy.cpp(include Lazy,h) and the other header file both do it
> 
> Include *.cpp in header looks more bad to me...

not include cpp to header，add new cpp file， and include the related header

eg:
add Lazy.cpp with context

```cpp
#include <Lazy.h>
```

ChuanqiXu9: > Is it possible to make coro part header only and standalone? that will be very easy to use for stackless coroutine scene.

It should be OK. The lib here is a fake one.

> > > you mean completely one empty file?
> > 
> > 
> > Yeah, the semantics look more clear to me.
> > > i mean Lazy.h we add Lazy.cpp(include Lazy,h) and the other header file both do it
> > 
> > 
> > Include *.cpp in header looks more bad to me...
> 
> not include cpp to header，add new cpp file， and include the related header
> 
> eg: add Lazy.cpp with context
> 
> ```c++
> #include <Lazy.h>
> ```

I am not sure why this is better than an empty file? A empty file is simpler.

RainMark: both not better, i thought you meant that before

RainMark: empty file too hack, how about to try cmake INTERFACE to create a header-only target

qicosmos: I like header only,haha

ChuanqiXu9: I think the current problem is that we need to generate library files so that both we and user could write build scripts uniformly instead of checking existence of UTHREAD ahead all the time. So given the solution, a header-only target might not be suitable. 

My thinking is that, if we need library files, we must need one source file at least. Then if we need a source file, what should we write? Nothing now. If we need nothing and a source file, then an empty source file makes sense. In this way , I think an empty file is a natural way to solve the problem.

RainMark: > I think the current problem is that we need to generate library files so that both we and user could write build scripts uniformly instead of checking existence of UTHREAD ahead all the time. So given the solution, a header-only target might not be suitable.
> 
> My thinking is that, if we need library files, we must need one source file at least. Then if we need a source file, what should we write? Nothing now. If we need nothing and a source file, then an empty source file makes sense. In this way , I think an empty file is a natural way to solve the problem.

agree, but i prefer to create source file for Condition.h or SimpleExecutor.h an so on.
if use empty source file, the others have to remember remove the empty file once we use a actual source file, leaved a hide problem for others

ChuanqiXu9: We could add a FIXME in the CMakeLists.txt so that others who edit it wouldn't forget it. Condition.cpp or SimpleExecutor.cpp is acceptable too. I just worried that it might hurt the performance before. But given SimpleExecutor is used in test only by design. It should be OK to generate SimpleExecutor.cpp.

RainMark: > We could add a FIXME in the CMakeLists.txt so that others who edit it wouldn't forget it. Condition.cpp or SimpleExecutor.cpp is acceptable too. I just worried that it might hurt the performance before. But given SimpleExecutor is used in test only by design. It should be OK to generate SimpleExecutor.cpp.

em, so we add the SimpleExecutor.cpp ?

ChuanqiXu9: Yeah, I agree

ChuanqiXu9 (approved):

RainMark (approved):

PR #74, opened 2022-03-24
Replace union to variant

### LINKED ISSUE #59: Undefined Behavior to access non active union member
### Search before asking

Extract from: https://www.reddit.com/r/cpp/comments/ta4h29/async_simple_a_c20_coroutine_library/:

```
ContextUnion u;
u.ctx = ctx;

auto prompt =
    _pool.getCurrentId() == (u.id & 0xBFFFFFFF) && opts.prompt;
```

It is undefined behavior, access to non active union member

qicosmos:
## Why

Use std::variant avoid undefined behavior of visit inactive union member.
For Close #59

### DIFF
--- a/async_simple/Executor.h
+++ b/async_simple/Executor.h
@@ -23,6 +23,7 @@
 #include <functional>
 #include <string>
 #include <thread>
+#include <variant>
 
 namespace async_simple {
 // Stat information for an executor.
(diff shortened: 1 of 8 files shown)


ChuanqiXu9:
I think this is not intended for #59. Since Executor::Context is not union at all. So this code is not consistent with #59. And #59 is closed after all.

For the patch it self, I feel it is not good. Since it limits the type of Context of Executor in the base class. And it would limit the use of Context in the derived class. Although the original style is in C style and the patch is more C++ style, the original style is more extendable in this case.

To make Executor::Context more C++ style, one solution might be define Context as a class. And we need to design a proxy and some interfaces for it. So that the user could define their wanted context. But I feel this is relatively over-designed. The current one is not bad though.

qicosmos:
Oh, i don't notice the ContextUnion has been removed, so there is no UB any more, it's ok, i will close this PR.

PR #75, opened 2022-03-25
[Cross Platform]Refactor thread pool

qicosmos:
## Why

The current thread pool don't support other platform, refactor it to provide a cross-platform, work-steal thread pool.


qicosmos:
Based on another PR, i will rebase main soon.

PR #76, opened 2022-03-25
[Cross Platform]Refactor thread pool

qicosmos:
## Why

The current thread pool don't support cross platform, refactor it to provide a cross-platform, work-steal thread pool.

### DIFF
--- a/async_simple/executors/SimpleExecutor.cpp
+++ b/async_simple/executors/SimpleExecutor.cpp
@@ -23,7 +23,6 @@ namespace executors {
 static constexpr int64_t kContextMask = 0x40000000;
 
 SimpleExecutor::SimpleExecutor(size_t threadNum) : _pool(threadNum) {
-    assert(ret);
     _ioExecutor.init();
 }
 
--- a/async_simple/executors/SimpleExecutor.cpp
+++ b/async_simple/executors/SimpleExecutor.cpp
@@ -23,15 +23,11 @@ namespace executors {
 static constexpr int64_t kContextMask = 0x40000000;
 
 SimpleExecutor::SimpleExecutor(size_t threadNum) : _pool(threadNum) {
-    [[maybe_unused]] auto ret = _pool.start();
     assert(ret);
     _ioExecutor.init();
 }
 
-SimpleExecutor::~SimpleExecutor() {
-    _ioExecutor.destroy();
-    _pool.stop();
-}
+SimpleExecutor::~SimpleExecutor() { _ioExecutor.destroy(); }
 
 SimpleExecutor::Context SimpleExecutor::checkout() {
     // avoid CurrentId equal to NULLCTX
--- a/async_simple/util/Queue.h
+++ b/async_simple/util/Queue.h
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2022, Alibaba Group Holding Limited;
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* The file implements a simple thread pool. The scheduling strategy is a simple
+ * random strategy. Simply, ThreadPool would create n threads. And for a task
+ * which is waiting to be scheduled. The ThreadPool would choose a thread for
+ * this task randomly.
+ *
+ * The purpose of ThreadPool is for testing. People who want to use async_simple
+ * in actual development should implement/use more complex executor.
+ */
+#ifndef ASYNC_SIMPLE_QUEUE_H
+#define ASYNC_SIMPLE_QUEUE_H
+
+#include <condition_variable>
+#include <mutex>
+#include <queue>
+#include <utility>
+
+namespace async_simple::util {
+
+template <typename T>
+class Queue {
+public:
+    Queue() = default;
+
+    void push(T &&item) {
+        {
+            std::scoped_lock guard(_mutex);
+            _queue.push(std::move(item));
+        }
+        _cond.notify_one();
+    }
+
+    bool try_push(const T &item) {
+        {
+            std::unique_lock lock(_mutex, std::try_to_lock);
+            if (!lock)
+                return false;
+            _queue.push(item);
+        }
+        _cond.notify_one();
+        return true;
+    }
+
+    bool pop(T &item) {
+        std::unique_lock lock(_mutex);
+        _cond.wait(lock, [&]() { return !_queue.empty() || _stop; });
+        if (_queue.empty())
+            return false;
+        item = std::move(_queue.front());
+        _queue.pop();
+        return true;
+    }
+
+    // non-blocking pop an item, maybe pop failed.
+    // predict is an extension pop condition, default is null.
+    bool try_pop(T &item, bool (*predict)(T &) = nullptr) {
+        std::unique_lock lock(_mutex, std::try_to_lock);
+        if (!lock || _queue.empty())
+            return false;
+
+        if (predict && predict(_queue.front())) {
+            return false;
+        }
+
+        item = std::move(_queue.front());
+        _queue.pop();
+        return true;
+    }
+
+    std::size_t size() const {
+        std::scoped_lock guard(_mutex);
+        return _queue.size();
+    }
+
+    bool empty() const {
+        std::scoped_lock guard(_mutex);
+        return _queue.empty();
+    }
+
+    void stop() {
+        {
+            std::scoped_lock guard(_mutex);
+            _stop = true;
+        }
+        _cond.notify_all();
+    }
+
+private:
+    std::queue<T> _queue;
+    bool _stop = false;
+    mutable std::mutex _mutex;
+    std::condition_variable _cond;
+};
+}  // namespace async_simple::util
+#endif  // ASYNC_SIMPLE_QUEUE_H
--- a/async_simple/util/Queue.h
+++ b/async_simple/util/Queue.h
@@ -32,6 +32,7 @@
 namespace async_simple::util {
 
 template <typename T>
+requires std::is_move_assignable_v<T>
 class Queue {
 public:
     void push(T &&item) {
--- a/async_simple/util/Queue.h
+++ b/async_simple/util/Queue.h
@@ -32,8 +32,7 @@
 namespace async_simple::util {
 
 template <typename T>
-requires std::is_move_assignable_v<T>
-class Queue {
+requires std::is_move_assignable_v<T> class Queue {
 public:
     void push(T &&item) {
         {
--- a/async_simple/util/Queue.h
+++ b/async_simple/util/Queue.h
@@ -34,8 +34,6 @@ namespace async_simple::util {
 template <typename T>
 class Queue {
 public:
-    Queue() = default;
-
     void push(T &&item) {
         {
             std::scoped_lock guard(_mutex);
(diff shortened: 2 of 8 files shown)


async_simple/util/ThreadPool.h
```diff
@@ -25,372 +25,137 @@
...
+        // If the user don't assign a thread to fn,
+        // thread pool will apply random policy to fn.
+        // If enable work steal,
+        // thread pool will apply work steal policy firstly , if failed, will
+        // apply random policy to fn.
+        bool autoSchedule = false;
+        std::function<void()> fn = nullptr;
     };
-
+    static constexpr size_t DEFAULT_THREAD_NUM = 4;
```
ChuanqiXu9, line 46: The name style of `DEFAULT_THREAD_NUM` looks like macro. And we should avoid use of magic number. We could use `std::thread::hardware_concurrency()` by default.

async_simple/util/Queue.h
```diff
@@ -0,0 +1,109 @@
...
+#include <queue>
+#include <utility>
+
+namespace async_simple::util {
+
+template <typename T>
+class Queue {
+public:
+    Queue() = default;
+
```
ChuanqiXu9, line 38: From the implementation, I think we should add a static_assert: static_assert(std:: is_move_assignable_v<T>, xxxx);

async_simple/executors/SimpleExecutor.cpp
```diff
@@ -23,15 +23,11 @@ namespace executors {
...
-    [[maybe_unused]] auto ret = _pool.start();
     assert(ret);
     _ioExecutor.init();
 }
 
-SimpleExecutor::~SimpleExecutor() {
-    _ioExecutor.destroy();
-    _pool.stop();
-}
+SimpleExecutor::~SimpleExecutor() { _ioExecutor.destroy(); }
```
RainMark, line 30: can guarantee all pushed task be executed before SimpleExecutor destroyed

async_simple/executors/SimpleExecutor.cpp
```diff
@@ -23,15 +23,11 @@ namespace executors {
 static constexpr int64_t kContextMask = 0x40000000;
 
 SimpleExecutor::SimpleExecutor(size_t threadNum) : _pool(threadNum) {
-    [[maybe_unused]] auto ret = _pool.start();
     assert(ret);
```
RainMark, line 26: have you tested on debug mode?

async_simple/util/ThreadPool.h
```diff
@@ -25,372 +25,137 @@
...
-public:
-    void process() { fn(); }
-    void destroy() { delete this; }
-    void drop() { delete this; }
-
-private:
-    std::function<void()> fn;
-};
-
+#include "Queue.h"
```
RainMark, line 32: i prefer <async_simple/util/Queue.h>

async_simple/util/test/ThreadPoolTest.cpp
```diff
@@ -68,4 +68,74 @@ TEST_F(ThreadPoolTest, testScheduleWithId) {
...
+
+private:
+    const char* m_name;
+    std::chrono::time_point<std::chrono::high_resolution_clock> m_beg;
+};
+
+const unsigned int COUNT = 1'000'000;
+const unsigned int REPS = 10;
+
+void Benchmark(bool enableWorkSteal) {
```
RainMark, line 115: how about to use benchmark framework

async_simple/util/Queue.h
```diff
@@ -0,0 +1,109 @@
...
+        if (_queue.empty())
+            return false;
+        item = std::move(_queue.front());
+        _queue.pop();
+        return true;
+    }
+
+    // non-blocking pop an item, maybe pop failed.
+    // predict is an extension pop condition, default is null.
+    bool try_pop(T &item, bool (*predict)(T &) = nullptr) {
```
ChuanqiXu9, line 70: I feel like `try_pop` carries too many functionalities. From the name `try_pop` shouldn't take a predicate. If we want to pop takes a predicate, we should name it `pop_if`. And if we want to make `try_pop` takes a predicate, we should name it `try_pop_if`. If all of this is needed, let's split it into the 3.

async_simple/util/ThreadPool.h
```diff
@@ -25,372 +25,137 @@
...
-    if (_threadNum == 0) {
-        _threadNum = DEFAULT_THREADNUM;
-    }
-}
+            if (!workerItem.fn && !_queues[id].pop(workerItem)) {
+                break;
+            }
 
-inline ThreadPool::~ThreadPool() { stop(STOP_AND_CLEAR_QUEUE); }
+            if (workerItem.fn) {
```
ChuanqiXu9, line 94: We could eliminate such braces.

async_simple/util/test/ThreadPoolTest.cpp
ChuanqiXu9, line 115: Let's add benchmarks in benchmarks directory

ChuanqiXu9 (commented):
We should add unittest for queue.

async_simple/util/ThreadPool.h
```diff
@@ -25,372 +25,137 @@
...
-        STOP_AND_CLEAR_QUEUE,
-        STOP_AFTER_QUEUE_EMPTY
+    struct WorkItem {
+        // Whether or not the fn is auto schedule.
+        // If the user don't assign a thread to fn,
+        // thread pool will apply random policy to fn.
+        // If enable work steal,
+        // thread pool will apply work steal policy firstly , if failed, will
+        // apply random policy to fn.
+        bool autoSchedule = false;
```
RainMark, line 43: the name of autoSchedule make me confused, you mean this item can be steal?

async_simple/util/ThreadPool.h
```diff
@@ -25,372 +25,137 @@
...
+        thread.join();
 }
 
-inline bool ThreadPool::start() {
-    if (_run) {
-        return false;
+inline ThreadPool::ERROR_TYPE ThreadPool::scheduleById(std::function<void()> fn,
+                                                       size_t id) {
+    if (nullptr == fn) {
+        return ERROR_POOL_ITEM_IS_NULL;
```
RainMark, line 116: ERROR_INVALID_ARGUMENTS?

async_simple/executors/SimpleExecutor.cpp
qicosmos: Yes, will drain off all the tasks when the pool destructed.

async_simple/executors/SimpleExecutor.cpp
qicosmos: fix a bug on debug mode, now tests on debug mode and release mode passed.

async_simple/util/test/ThreadPoolTest.cpp
qicosmos: OK

async_simple/util/ThreadPool.h
qicosmos: OK

async_simple/util/ThreadPool.h
qicosmos: If the user don't set a scheduling thread id, the autoSchedule(if it's value is true) will choose any thread  to execute, otherwise assigned the task to the given thread.

steal is another attitude, if autoSchedule is true and steal flag is true, the work steal will be trigger;
if schedule with a thread id, work steal will never trigger.

async_simple/util/test/ThreadPoolTest.cpp
qicosmos: ok

async_simple/util/Queue.h
qicosmos: sounds good.

async_simple/util/ThreadPool.h
qicosmos: If the user pass nullptr to threadpool workerItem.fn() will crash.
Do you mean use ```assert(workerItem.fn);``` instead of null checking?

async_simple/util/Queue.h
qicosmos: I think the ```Queue() = default;``` can be removed, because of it has mutex and it is non-copyable.

async_simple/util/ThreadPool.h
```diff
@@ -26,325 +26,130 @@
...
-
-inline void ThreadPool::stop(STOP_TYPE stopType) {
-    if (STOP_THREAD_ONLY != stopType) {
-        { _push = false; }
-    }
-
-    if (STOP_AFTER_QUEUE_EMPTY == stopType) {
-        waitQueueEmpty();
+        assert(id < _threadNum);
+        _queues[id].push(WorkItem{true, std::move(fn)});
```
RainMark, line 142: WorkItem{true means this task can be steal by other thread?

async_simple/util/ThreadPool.h
RainMark: how about to rename autoSchedule to stealable/canSteal/migratable or transferable and so on

qicosmos: Right, canSteal will make the work steal syntax more clear.

async_simple/util/Queue.h
```diff
@@ -0,0 +1,117 @@
...
+                return false;
+            _queue.push(item);
+        }
+        _cond.notify_one();
+        return true;
+    }
+
+    bool pop(T &item) {
+        std::unique_lock lock(_mutex);
+        _cond.wait(lock, [&]() { return !_queue.empty() || _stop; });
```
ChuanqiXu9, line 58: Why do we need conditional variable here?

async_simple/util/Queue.h
```diff
@@ -0,0 +1,117 @@
...
+        _queue.pop();
+        return true;
+    }
+
+    bool try_pop(T &item) {
+        std::unique_lock lock(_mutex, std::try_to_lock);
+        if (!lock || _queue.empty())
+            return false;
+
+        item = std::move(_queue.front());
```
ChuanqiXu9, line 71: Given we call T::operator =(T&&) here, it would better to add static_assert(std::is_move_assignable_v<T>) in the class.

async_simple/util/ThreadPool.h
qicosmos: Do a improvement:
```
    if (id == -1) {
        if (_enableWorkSteal) {
            // Try to push to a non-block queue firstly.
            WorkItem workerItem{/*canSteal = */ true, fn};
            for (auto n = 0; n < _threadNum * 2; ++n) {
                if (_queues.at(n % _threadNum).try_push(workerItem))
                    return ERROR_NONE;
            }
        }

        id = rand() % _threadNum;
        _queues[id].push(
            WorkItem{/*canSteal = */ _enableWorkSteal, std::move(fn)});
    } else {
        assert(id < _threadNum);
        _queues[id].push(WorkItem{/*canSteal = */ false, std::move(fn)});
    }
```
id is -1 and _enableWorkSteal true: can worksteal;
otherwise: never worksteal;

Is this code better and clear?

async_simple/util/Queue.h
qicosmos: pop() wait until get a task, if the queue is empty, will blocking waiting.

async_simple/util/Queue.h
qicosmos: ok

async_simple/util/ThreadPool.h
RainMark: 👍🏻  nice

async_simple/util/ThreadPool.h
```diff
@@ -26,325 +26,133 @@
...
-    enum {
-        DEFAULT_THREADNUM = 4,
-    };
-
-    enum STOP_TYPE {
-        STOP_THREAD_ONLY = 0,
-        STOP_AND_CLEAR_QUEUE,
-        STOP_AFTER_QUEUE_EMPTY
+    struct WorkItem {
+        // Whether or not the fn is auto schedule.
```
RainMark, line 38: - the comment need update

qicosmos: OK, has updated the comments.

RainMark:
LGTM

RainMark (approved):

ChuanqiXu9 (approved):

Issue #17, opened 2022-02-22
lazy.start() captured exception by default


RainMark:
### Search before asking

- [ ] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

### What happened + What you expected to happen

lazy.start() captured exception by default and pass Try to user callback
user maybe forget to process exception, caused user's coroutine code can not finish, and the resource leaked.

make lazy.start() throw exception by default?

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

RainMark:
if user do not want to handle exception, they can use `detach()` , so we let user to decided by self

PR #78, opened 2022-03-28
[Simplify Interface]add detach

qicosmos:
## Why
simplify code:
```via(&executor).start([](auto&&) {});```
 to
```via(&executor).detach();```

### DIFF
--- a/async_simple/coro/Lazy.h
+++ b/async_simple/coro/Lazy.h
@@ -510,7 +510,11 @@ public:
     }
 
     void detach() {
-        start([](auto&&) {});
+        start([](auto&& t) {
+            if (t.hasError()) {
+                std::rethrow_exception(t.getException());
+            }
+        });
     }
 
 private:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: async_simple/coro/Lazy.h [hunk 1]
```diff
@@ -509,6 +509,10 @@ class [[nodiscard]] RescheduleLazy {
             launchCoro(std::move(*this), std::forward<F>(callback));
     }
 
+    void detach() {
+        start([](auto&&) {});
```
RainMark, line 513: in previous issue #17 we know user forget to handle exception with `start([](auto&&) {});`
we have to rethrow the existing exception in `detach` api, because user can not use Try type to handle it any more

```cpp
start([](auto&& t) {
  if (t.hasError()) {
    // rethrow
  }
});
```

qicosmos: OK

ChuanqiXu9 (approved):

RainMark (approved):

PR #83, opened 2022-03-29
Enable to use coroutine components under std directly

ChuanqiXu9:
We could use coroutine components under std directly by using
experimental ones in std namespace. So the code would be cleaner.


qicosmos (approved):
Great, get rid of macro at last:)

PR #84, opened 2022-03-31
Support windows

### LINKED ISSUE #72: [Cross Platform]Support Windows
### Search before asking

We have supported Mac, the next is Windows, at least make sure supporting Windows when using c++20 coroutine.

- [x] Yes I am willing to submit a PR!

qicosmos:
## Why

Closes #72 

## What is changing

name conflicts to windows:
- _try --> _try_value
- Yield() --> Yield_t()

not exist functions on windows:
- ```__builtin_expect --> #define __builtin_expect(EXP, C) (EXP)```
- ```__attribute__((__always_inline__)) --> #define FL_INLINE inline```

change some compile options for windows:
- pthread
- async_simple static lib
- some c++ compile options

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ endif()
 set(testdeplibs)
 list(APPEND testdeplibs ${GTEST_LIBRARIES})
 list(APPEND testdeplibs ${GMOCK_LIBRARIES})
+list(APPEND testdeplibs ${CMAKE_THREAD_LIBS_INIT})
 
 if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
     if(CMAKE_BUILD_BITS EQUAL 32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ enable_testing()
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 message(STATUS "CMAKE_MODULE_PATH: ${CMAKE_MODULE_PATH}")
 
+find_package(Threads)
 find_package(Aio)
 find_package(GMock REQUIRED)
 find_package(GTest REQUIRED)
@@ -42,17 +43,25 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 # set CXXFALGS
-set(CXX_FLAGS
-    -std=c++20
-    -D_GLIBCXX_USE_CXX11_ABI=1
-    -Wno-deprecated-register
-    -D_FILE_OFFSET_BITS=64
-    -fPIC
-    -Wall
-    -Werror
-    -D__STDC_LIMIT_MACROS
-    -g
-    )
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    set(CXX_FLAGS
+        /std:c++20
+        /await:strict
+        /EHa
+        )
+else()
+    set(CXX_FLAGS
+        -std=c++20
+        -D_GLIBCXX_USE_CXX11_ABI=1
+        -Wno-deprecated-register
+        -D_FILE_OFFSET_BITS=64
+        -fPIC
+        -Wall
+        -Werror
+        -D__STDC_LIMIT_MACROS
+        -g
+        )
+endif()
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     list(APPEND CXX_FLAGS "-fcoroutines")
@@ -67,24 +76,27 @@ set(HEADERS_PATH
 include_directories(${HEADERS_PATH})
 
 set(deplibs)
+list(APPEND deplibs ${CMAKE_THREAD_LIBS_INIT})
 if(LIBAIO_LIBRARIES)
     list(APPEND deplibs ${LIBAIO_LIBRARIES})
 endif()
-list(APPEND deplibs pthread)
+
 set(testdeplibs)
 list(APPEND testdeplibs ${GTEST_LIBRARIES})
 list(APPEND testdeplibs ${GMOCK_LIBRARIES})
 
-if(CMAKE_BUILD_BITS EQUAL 32)
-    message("-- Use flag -m32")
-    list(APPEND CXX_FLAGS "-m32")
-    list(APPEND CXX_FLAGS "-DTARGET_32")
-    list(APPEND deplibs "-m32")
-else()
-    message("-- Use flag -m64")
-    list(APPEND CXX_FLAGS "-m64")
-    list(APPEND CXX_FLAGS "-DTARGET_64")
-    #list(APPEND deplibs "-m64")
+if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    if(CMAKE_BUILD_BITS EQUAL 32)
+        message("-- Use flag -m32")
+        list(APPEND CXX_FLAGS "-m32")
+        list(APPEND CXX_FLAGS "-DTARGET_32")
+        list(APPEND deplibs "-m32")
+    else()
+        message("-- Use flag -m64")
+        list(APPEND CXX_FLAGS "-m64")
+        list(APPEND CXX_FLAGS "-DTARGET_64")
+        #list(APPEND deplibs "-m64")
+    endif()
 endif()
 
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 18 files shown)


CMakeLists.txt
```diff
@@ -67,24 +76,27 @@ set(HEADERS_PATH
...
+        list(APPEND deplibs "-m32")
+    else()
+        message("-- Use flag -m64")
+        list(APPEND CXX_FLAGS "-m64")
+        list(APPEND CXX_FLAGS "-DTARGET_64")
+        #list(APPEND deplibs "-m64")
+    endif()
 endif()
 
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
```
ChuanqiXu9, line 102: I guess `-fsanitize=address` is not meaningful in windows.

async_simple/CMakeLists.txt
```diff
@@ -30,15 +30,15 @@ if(UTHREAD)
   list(APPEND SRCS ${uthread_asm_src})
 endif()
 
-add_library(async_simple_static STATIC ${SRCS})
-add_library(async_simple SHARED ${SRCS})
+add_library(async_simple STATIC ${SRCS})
```
ChuanqiXu9, line 33: Is this change necessary?

async_simple/Common.h
```diff
@@ -18,10 +18,18 @@
 
 #include <stdexcept>
 
+#ifdef _WIN32
+#define __builtin_expect(EXP, C) (EXP)
+#endif
+
 #define FL_LIKELY(x) __builtin_expect((x), 1)
 #define FL_UNLIKELY(x) __builtin_expect((x), 0)
```
ChuanqiXu9, line 26: We could use `[[likely]]` one day.

async_simple/Common.h
```diff
@@ -18,10 +18,18 @@
...
 
+#ifdef _WIN32
+#define FL_INLINE inline
```
ChuanqiXu9, line 29: Better rename `FL_INLINE` to `AS_INLNIE`

async_simple/FutureState.h
```diff
@@ -168,8 +168,8 @@ class FutureState {
     }
 
 public:
-    Try<T>& getTry() noexcept { return _try; }
-    const Try<T>& getTry() const noexcept { return _try; }
+    Try<T>& getTry() noexcept { return _try_value; }
```
ChuanqiXu9, line 171: Why the renaming is necessary?

async_simple/coro/Lazy.h
```diff
@@ -43,7 +43,7 @@ class Lazy;
 // ```
 //
 // This would suspend the executing coruotine.
-struct Yield {};
+struct Yield_t {};
```
ChuanqiXu9, line 46: Is this necessary? Why?

async_simple/experimental/coroutine.h
```diff
@@ -55,7 +55,8 @@ template <class P> struct hash<coroutine_handle<P>>;
 // we could only use self-provided coroutine header.
 // Note: the <coroutine> header in libc++ is available for both clang and gcc.
 // And the outdated <experimental/coroutine> is available for clang only.
-#if (__cplusplus <= 201703L) || (defined(__clang__) && defined(__GLIBCXX__))
+#if (__cplusplus <= 201703L && !defined(_MSC_VER)) || \
```
ChuanqiXu9, line 58: Edit the comment too.

async_simple/executors/SimpleIOExecutor.h
```diff
@@ -20,7 +20,9 @@
 #ifndef ASYNC_SIMPLE_HAS_NOT_AIO
 #include <libaio.h>
 #endif
+#ifndef _WIN32
 #include <sys/syscall.h> /* For SYS_xxx definitions */
```
RainMark, line 24: this header can be removed directly？

qicosmos:
I will have a try and show more error messages after testing on Windows later.

async_simple/coro/Lazy.h
qicosmos: Has reverted, it's related with asio.hpp header sequence.

async_simple/FutureState.h
qicosmos: Can't use _try or __try variable in windows, will cause confilict:https://godbolt.org/z/KcYMsTvx5

async_simple/Common.h
qicosmos: Actually msvc, gcc and clang12+ supported likely/unlikely, but ci clang version is clang11, will have a unknown-attributes warning. Maybe add a compile flag to avoid this warning errorr.

async_simple/coro/Lazy.h
qicosmos: Can be removed.

async_simple/executors/SimpleIOExecutor.h
qicosmos: Useless, can be removed.

async_simple/CMakeLists.txt
qicosmos: If use the original config, will generate async_imple.dll on windows, and all the test cases depends on the async_simple.lib and compile error on windows, so here don't use set_target_properties. Explicit generate static and shared lib.

async_simple/Common.h
```diff
@@ -18,10 +18,15 @@
 
 #include <stdexcept>
 
-#define FL_LIKELY(x) __builtin_expect((x), 1)
-#define FL_UNLIKELY(x) __builtin_expect((x), 0)
+#ifdef _WIN32
+#define __builtin_expect(EXP, C) (EXP)
+#endif
```
ChuanqiXu9, line 23: Is this still useful?

async_simple/Common.h
```diff
@@ -18,10 +18,15 @@
...
 
-#define FL_INLINE __attribute__((__always_inline__)) inline
+#ifdef _WIN32
+#define AS_INLINE inline
```
ChuanqiXu9, line 26: `inline` should have other meaning than suggesting a inlining. Let's define AS_INLINE as empty for windows.

async_simple/CMakeLists.txt
ChuanqiXu9: Now the behavior changes more. Currently, the shared library wouldn't get installed. Also the name changes. Both is not good. I would suggest to wrap line 38 for windows only.

async_simple/Common.h
```diff
@@ -31,9 +36,14 @@ namespace async_simple {
...
 inline void logicAssert(bool x, const char* errorMsg) {
-    if (FL_LIKELY(x)) {
+    if (x)
+#if defined(__clang__) && __clang_major__ < 12
+    {
         return;
     }
+#else
+        [[likely]] { return; }
+#endif
```
ChuanqiXu9, line 46: This looks redundant. I would suggest:
```suggestion
    if (x) AS_LIKELY
        return;
```

async_simple/executors/test/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 file(GLOB executor_test_src "*.cpp")
 add_executable(async_simple_executor_test ${executor_test_src} ${PROJECT_SOURCE_DIR}/async_simple/test/dotest.cpp)
 
-target_link_libraries(async_simple_executor_test async_simple -pthread ${deplibs} ${testdeplibs})
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    target_link_libraries(async_simple_executor_test async_simple ${deplibs} ${testdeplibs})
+else()
+    target_link_libraries(async_simple_executor_test async_simple -pthread ${deplibs} ${testdeplibs})
+endif()
```
ChuanqiXu9, line 8: From the context, it looks like we could use ${CMAKE_THREAD_LIBS_INIT} here?

ChuanqiXu9 (changes requested):

async_simple/executors/test/CMakeLists.txt
qicosmos: No need change, revert is ok on windows, maybe i did not compile all projects cause such error before.

async_simple/Common.h
qicosmos: Good

async_simple/Common.h
qicosmos: OK

async_simple/Common.h
qicosmos: Removed

ChuanqiXu9 (approved):
The formatted code for `LIKELY` macro looks ugly. But it looks like we couldn't do any thing better. We need update documents and CI for windows too. Let's do it in another PR.

PR #93, opened 2022-04-15
Don't set std::hash<coroutine> in case we use std implementation

ChuanqiXu9:
## Why

If we chose standard implementation for coroutine, the current implementation might met a redefinition error for std::hash<std::coroutine> type.


qicosmos (approved):
LGTM

PR #96, opened 2022-04-24
[coro] add spin lock

RainMark:
## Why

add spin lock implement based on c++20 coroutines

## What is changing

new feature

### DIFF
--- a/async_simple/coro/SpinLock.h
+++ b/async_simple/coro/SpinLock.h
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2022, Alibaba Group Holding Limited;
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


ChuanqiXu9:
Would you like to add a simple document for its purpose and usage?

RainMark:
> Would you like to add a simple document for its purpose and usage

> Would you like to add a simple document for its purpose and usage?

After condition variable added, i will add documents and demo example

async_simple/coro/SpinLock.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include <async_simple/coro/Lazy.h>
+#include <thread>
+
+namespace async_simple {
+namespace coro {
+
+class SpinLock {
+public:
+    explicit SpinLock(std::int32_t count = 1024) noexcept
```
qicosmos, line 27: I see some other coroutine spinlock don't have a count argument:
[quantum_spinlock](https://github.com/bloomberg/quantum/blob/master/quantum/quantum_spinlock.h), do we need this count?

RainMark: use spin count to avoid current coroutine occupy current thread to long, and starve the others coroutines.

I checked [quantum_spinlock_util_impl](https://github.com/bloomberg/quantum/blob/master/quantum/util/impl/quantum_spinlock_util_impl.h), it can not yield current coroutine when not get the lock.

ChuanqiXu9: I feel good to offer a count to make the user to control the behavior. quantum_spinlock is used internally so we have different cases.

ChuanqiXu9:
> > Would you like to add a simple document for its purpose and usage
> 
> > Would you like to add a simple document for its purpose and usage?
> 
> After condition variable added, i will add documents and demo example

OK, it looks good to me.

ChuanqiXu9 (approved):

qicosmos (approved):
LGTM

Issue #92, opened 2022-04-13
make test receive that error!


moruoxian:
### Search before asking

- [x] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

### What happened + What you expected to happen

root@win10:/home/Project/async_simple/build# make test
Running tests...
Test project /home/Project/async_simple/build
    Start 1: run_async_simple_test
1/5 Test #1: run_async_simple_test ............   Passed    2.73 sec
    Start 2: run_async_simple_util_test
2/5 Test #2: run_async_simple_util_test .......   Passed    0.02 sec
    Start 3: run_async_simple_coro_test
3/5 Test #3: run_async_simple_coro_test .......***Failed   37.21 sec
    Start 4: run_async_simple_executor_test
4/5 Test #4: run_async_simple_executor_test ...   Passed    1.07 sec
    Start 5: run_async_simple_uthread_test
5/5 Test #5: run_async_simple_uthread_test ....   Passed    3.95 sec

80% tests passed, 1 tests failed out of 5

Total Test time (real) =  44.99 sec

The following tests FAILED:
          3 - run_async_simple_coro_test (Failed)
Errors while running CTest

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

moruoxian:
ubuntu 20.04
Linux win10 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
gcc version 10.3.0 (Ubuntu 10.3.0-1ubuntu1~20.04)

qicosmos:
I tested on my ububtu20.04, gcc11.1

```
Test project /home/ubuntu/code/async_simple/build
    Start 1: run_async_simple_test
1/5 Test #1: run_async_simple_test ............   Passed    1.21 sec
    Start 2: run_async_simple_util_test
2/5 Test #2: run_async_simple_util_test .......   Passed    0.00 sec
    Start 3: run_async_simple_coro_test
3/5 Test #3: run_async_simple_coro_test .......   Passed   12.81 sec
    Start 4: run_async_simple_executor_test
4/5 Test #4: run_async_simple_executor_test ...   Passed    0.00 sec
    Start 5: run_async_simple_uthread_test
5/5 Test #5: run_async_simple_uthread_test ....   Passed    1.66 sec

100% tests passed, 0 tests failed out of 5

[...]

PR #97, opened 2022-04-25
Decrease the scale in testLazyPerf to make GCC compatible

ChuanqiXu9:
## Why

Closing #92 

The reason here should be the lackness of the optimization of GCC for coroutines.

### DIFF
--- a/async_simple/coro/test/LazyTest.cpp
+++ b/async_simple/coro/test/LazyTest.cpp
@@ -1177,7 +1177,16 @@ Lazy<int> lazy_fn<0>() {
 }
 }  // namespace detail
 TEST_F(LazyTest, testLazyPerf) {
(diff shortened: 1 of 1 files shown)


qicosmos (approved):
LGTM

async_simple/coro/test/LazyTest.cpp
```diff
@@ -1177,7 +1177,16 @@ Lazy<int> lazy_fn<0>() {
 }
 }  // namespace detail
 TEST_F(LazyTest, testLazyPerf) {
```
RainMark, line 1179: we already have benchmark for performance test, this case can be removed?

ChuanqiXu9: More tests are never bad : )

PR #98, opened 2022-04-25
Remove unused using in unittest.h

ChuanqiXu9:
The using statement in unittest.h is meaningless and it is possible to
make async_simple incompatible for user whose version of goolgetest is
not high enough.


qicosmos (approved):
LGTM

PR #99, opened 2022-05-12
SpinLock.POC

beef9999:


qicosmos:
you'd better add a new example for testing spinlock, don't modify the current case.
And you need to sign our [Contributor License Agreement](https://cla-assistant.io/alibaba/async_simple?pullRequest=99) before we can accept your contribution.

beef9999:
... It's not a contribution. It's a demo to show the bug.

RainMark:
> ... It's not a contribution. It's a demo to show the bug.

Ok, we will try to test it, and can your supply the backtrace of segment fault?

ChuanqiXu9:
This might be a mis-use. We need to pass an executor in the start point of the async chain. But the diagnostic message from the library side is not friendly indeed. We would try to enhance it. Thanks for reporting.

demo_example/async_echo_client.cpp
```diff
@@ -17,9 +17,38 @@
...
+static async_simple::coro::SpinLock g_wlock, g_rlock;
 
 using asio::ip::tcp;
 
+async_simple::coro::Lazy<int> task(asio::ip::tcp::socket& socket) {
+    const int max_length = 1024;
+    char write_buf[max_length] = {"hello async_simple"};
+    char read_buf[max_length];
+    auto send = [&]() -> async_simple::coro::Lazy<> {
+        auto scopeLock = co_await g_wlock.coScopedLock();
```
RainMark, line 32: coScopedLock need a executor to reschedule current coroutine to avoid the others be blocked, but this demo code use syncAwait to start a lazy task (without executor), so segment fault occurred

you can use ScopedSpinLock to avoid the coroutine yield or pass a executor to syncAwait

beef9999:
I think what I need is a mutex(based on coroutine). I didn’t find it in the code.

beef9999:
And what kind of executor should I use there?

RainMark:
> I think what I need is a mutex(based on coroutine). I didn’t find it in the code.

ok, mutex is working on

RainMark:
> And what kind of executor should I use there?

just use async_simple::executors::SimpleExecutor for test

beef9999:
Is coScopedLock + executor sufficient to run my test? Because you said mutex is still under development … I’m a little confused

RainMark:
> Is coScopedLock + executor sufficient to run my test? Because you said mutex is still under development … I’m a little confused

yes

beef9999:
I have updated the code and added the Executor, now this test case is able to run.

But as long as I increase the `concurrency` (coroutine number), the QPS would continue to fall.

| Concurrency | QPS |
| --- | ----------- |
| 1 | 167K |
| 2 | 155K |
| 8 | 97K |
| 32 | 78K |

beef9999:
Looks like the coScopedLock + executor is not working as expected as a mutex does. Something is blocked.

demo_example/async_echo_client.cpp
```diff
@@ -29,28 +70,14 @@ async_simple::coro::Lazy<void> start(asio::io_context &io_context,
...
-            throw asio::system_error(error);
-        }
 
-        // handle read data as your wish.
+    std::vector<async_simple::coro::Lazy<>> input;
+    for (int i = 0; i < concurrency; ++i) {
+        input.push_back(task(socket));
     }
+    input.push_back(show_qps());
+    co_await async_simple::coro::collectAll(std::move(input));
```
RainMark, line 79: collectAll run multi coroutines in same thread asynchronously, looks like you want all coroutines parallel on all thread, try to use `collectAllPara`？

ChuanqiXu9:
> 

Thanks for reporting. I've got a similar results. And my conclusions are:
1. SpinLock is not performance friendly. When I remove SpinLock and use atomic counter, the QPS wouldn't decrease if we increase the conccurency.
2. SimpleExecutor is not good for performance too. We stated it is only used for testing in the docs. The direct impact would be: when we unlock a lock, we send the task to resuming a coroutine to the executor. But the executor wouldn't resume that immediately. The schedule strategy is simply random.

A condition variable here might be better.

ChuanqiXu9:
@beef9999 Given there is no other comment. I am going close this one. Thanks for contributing.

PR #102, opened 2022-05-13
[coro] add condition variable

RainMark:
## What is changing

add condition variable

note: the `TEST_F(ConditionVariableTest, testMultiWait)` can not passed on asan check. need fix it before merge

### DIFF
--- a/async_simple/coro/ConditionVariable.h
+++ b/async_simple/coro/ConditionVariable.h
@@ -24,7 +24,7 @@ namespace coro {
 template <class Lock>
 class ConditionVariableAwaiter;
 
-template <class Lock = void>
+template <class Lock>
 class ConditionVariable {
 public:
     ConditionVariable() noexcept {}
@@ -46,7 +46,7 @@ private:
     std::atomic<ConditionVariableAwaiter<Lock>*> _awaiters = nullptr;
 };
 
-template <class Lock = void>
+template <class Lock>
 class ConditionVariableAwaiter {
 public:
     ConditionVariableAwaiter(ConditionVariable<Lock>* cv, Lock& lock) noexcept
@@ -54,7 +54,7 @@ public:
 
     bool await_ready() const noexcept { return false; }
 
-    bool await_suspend(std::coroutine_handle<> continuation) noexcept {
+    void await_suspend(std::coroutine_handle<> continuation) noexcept {
         _continuation = continuation;
         std::unique_lock<Lock> lock{_lock, std::adopt_lock};
         auto awaitings = _cv->_awaiters.load(std::memory_order_relaxed);
@@ -63,7 +63,6 @@ public:
         } while (!_cv->_awaiters.compare_exchange_weak(
             awaitings, this, std::memory_order_acquire,
             std::memory_order_relaxed));
-        return true;
     }
     void await_resume() const noexcept {}
 
@@ -100,15 +99,17 @@ inline void ConditionVariable<Lock>::notify() noexcept {
 template <class Lock>
 inline void ConditionVariable<Lock>::resumeWaiters(
     ConditionVariableAwaiter<Lock>* awaiters) {
-    for (auto head = awaiters; head; head = head->_next) {
-        head->_continuation.resume();
+    while (awaiters) {
+        auto* prev = awaiters;
+        awaiters = awaiters->_next;
+        prev->_continuation.resume();
     }
 }
 
 template <>
-class ConditionVariableAwaiter<>;
+class ConditionVariableAwaiter<void>;
 template <>
-class ConditionVariable<> {
+class ConditionVariable<void> {
 public:
     using pointer_type = void*;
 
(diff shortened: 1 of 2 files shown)


ChuanqiXu9:
This contains spin lock. It looks redundant.

RainMark:
> This contains spin lock. It looks redundant.

already rebased

async_simple/coro/ConditionVariable.h
```diff
@@ -0,0 +1,200 @@
...
+inline void ConditionVariable<>::reset() noexcept {
+    pointer_type self = static_cast<pointer_type>(this);
+    _awaiters.compare_exchange_strong(self, nullptr, std::memory_order_relaxed);
+}
+
+inline void ConditionVariable<>::resumeWaiters(
+    ConditionVariableAwaiter<>* awaiters) {
+    for (auto head = awaiters; head; head = head->_next) {
+        head->_continuation.resume();
+    }
```
ChuanqiXu9, line 192:
```suggestion
    while (awaiters) {
        auto *prev = awaiters;
        awaiters = awaiters->_next;
        prev->_continuation.resume();
    }
```

We shouldn't run `head = head->_next` after `head->_continuation.resume();`. Since after the resumption, the coroutine the `head` refers might be released.

async_simple/coro/ConditionVariable.h
```diff
@@ -0,0 +1,200 @@
...
+    ConditionVariableAwaiter(ConditionVariable<>* cv) noexcept : _cv(cv) {}
+
+    bool await_ready() const noexcept {
+        return static_cast<pointer_type>(_cv) ==
+               _cv->_awaiters.load(std::memory_order_acquire);
+    }
+    bool await_suspend(std::coroutine_handle<> continuation) noexcept {
+        _continuation = continuation;
+        pointer_type awaitings = _cv->_awaiters.load(std::memory_order_acquire);
+        do {
```
ChuanqiXu9, line 149:
```suggestion
        pointer_type awaitings = nullptr;
        do {
             awaiting =  _cv->_awaiters.load(std::memory_order_acquire);
```

Otherwise, the `awaitings` wouldn't be updated in case the condition in while fails.

async_simple/coro/ConditionVariable.h
```diff
@@ -0,0 +1,200 @@
...
+
+#include <async_simple/coro/Lazy.h>
+
+namespace async_simple {
+namespace coro {
+
+template <class Lock>
+class ConditionVariableAwaiter;
+
+template <class Lock = void>
```
ChuanqiXu9, line 27: We should handle the `Lock = void` in case of `ConditionVariable<>`. Let's remove the default argument and add a static_assert.

async_simple/coro/ConditionVariable.h
```diff
@@ -0,0 +1,200 @@
...
+    ConditionVariableAwaiter(ConditionVariable<Lock>* cv, Lock& lock) noexcept
+        : _cv(cv), _lock(lock) {}
+
+    bool await_ready() const noexcept { return false; }
+
+    bool await_suspend(std::coroutine_handle<> continuation) noexcept {
+        _continuation = continuation;
+        std::unique_lock<Lock> lock{_lock, std::adopt_lock};
+        auto awaitings = _cv->_awaiters.load(std::memory_order_relaxed);
+        do {
```
ChuanqiXu9, line 61:
```suggestion
        void* awaitings = nullptr;
        do {
                awaiting = _cv->_awaiters.load(std::memory_order_relaxed);
```

async_simple/coro/ConditionVariable.h
```diff
@@ -0,0 +1,200 @@
...
+
+template <class Lock = void>
+class ConditionVariableAwaiter {
+public:
+    ConditionVariableAwaiter(ConditionVariable<Lock>* cv, Lock& lock) noexcept
+        : _cv(cv), _lock(lock) {}
+
+    bool await_ready() const noexcept { return false; }
+
+    bool await_suspend(std::coroutine_handle<> continuation) noexcept {
```
ChuanqiXu9, line 57: Since this would return `true` all the time. The return type could be `void`.

async_simple/coro/ConditionVariable.h
RainMark: apply this fix can pass the asan check？

async_simple/coro/ConditionVariable.h
RainMark: the `awaitings` will be updated to `_cv->_awaiters`'s latest value automatically when `cas` failed

async_simple/coro/ConditionVariable.h
RainMark: We already specialized the void type，its ok to remove `Lock = void`, and update to `using Notifier = ConditionVariable<void>;`

async_simple/coro/ConditionVariable.h
ChuanqiXu9: Oh, you are correct here.

ChuanqiXu9 (approved):
LGTM.

PR #104, opened 2022-05-16
[ThreadPool] Run a thraed per CPU in SimpleExecutor

ChuanqiXu9:
## Why

Now we don't bind threads in ThreadPool to physical CPU cores. So we might not be able to get the expected concurrency.

## What is changing

This one should be transparent for users.


async_simple/util/ThreadPool.h
```diff
@@ -93,19 +98,28 @@ inline ThreadPool::ThreadPool(size_t threadNum, bool enableWorkSteal)
...
 
     _threads.reserve(_threadNum);
     for (auto i = 0; i < _threadNum; ++i) {
         _threads.emplace_back(worker, i);
+
+#ifdef __linux__
+        // Run threads per core.
+        cpu_set_t cpuset;
+        CPU_ZERO(&cpuset);
+        CPU_SET(i, &cpuset);
```
RainMark, line 117: 1、cpu id maybe not start at `0` or be sequenced in container environment
2、`_threadNum` maybe great than available cpus
3、macOS support pthread？

async_simple/util/ThreadPool.h
```diff
@@ -93,19 +98,28 @@ inline ThreadPool::ThreadPool(size_t threadNum, bool enableWorkSteal)
...
+        int rc = pthread_setaffinity_np(_threads[i].native_handle(),
+                                        sizeof(cpu_set_t), &cpuset);
+        if (rc != 0)
+            std::cerr << "Error calling pthread_setaffinity_np: " << rc << "\n";
```
RainMark, line 121: we need a log component？

qicosmos (approved):
LGTM

qicosmos:
Do you compare the performance using this improvement? @ChuanqiXu9

async_simple/util/ThreadPool.h
ChuanqiXu9: Yeah, I filed an issue at https://github.com/alibaba/async_simple/issues/105

async_simple/util/ThreadPool.h
ChuanqiXu9: 1. According to https://linux.die.net/man/3/cpu_set, the cpu id starts at 0 and is continuous.
2. I refactored it to `CPU_SET(i % CPU_COUNT(i), &cpuset);` So it wouldn't overflow.
3. I didn't find related things in https://developer.apple.com/search/?q=CPU_SET. Given the users who would need performance in macOS should be less, I think we could give up to  support them.

ChuanqiXu9:
> Do you compare the performance using this improvement? @ChuanqiXu9

Yeah, I found this one when looking at https://github.com/alibaba/async_simple/pull/99

async_simple/util/ThreadPool.h
```diff
@@ -93,19 +115,41 @@ inline ThreadPool::ThreadPool(size_t threadNum, bool enableWorkSteal)
...
+
+#ifdef __linux__
+        if (!_enableCoreBindings)
+            continue;
+
+        // Run threads per core.
+        cpu_set_t cpuset;
+        CPU_ZERO(&cpuset);
+        CPU_SET(cpu_ids[i % cpu_ids.size()], &cpuset);
+        int rc = pthread_setaffinity_np(_threads[i].native_handle(),
```
RainMark, line 148: we can use  sched_setaffinity syscall to avoid include pthread，because already be linux system

RainMark:
LGTM

Issue #109, opened 2022-05-18
ThreadPool.h compile error


RainMark:
### Search before asking

- [ ] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

### What happened + What you expected to happen

```sh
$ make -j4
[  1%] Building CXX object async_simple/CMakeFiles/async_simple_static.dir/executors/SimpleExecutor.cpp.o
[  3%] Building CXX object async_simple/CMakeFiles/async_simple.dir/executors/SimpleExecutor.cpp.o
[  5%] Building CXX object demo_example/CMakeFiles/smtp_client.dir/smtp/smtp_client.cpp.o
[  7%] Building CXX object demo_example/CMakeFiles/block_echo_client.dir/block_echo_client.cpp.o
In file included from In file included from /Users/rain/Project/async_simple/async_simple/executors/SimpleExecutor.cpp:/Users/rain/Project/async_simple/async_simple/executors/SimpleExecutor.cpp16::
16In file included from :
/Users/rain/Project/async_simple/async_simple/executors/SimpleExecutor.hIn file included from :/Users/rain/Project/async_simple/async_simple/executors/SimpleExecutor.h23::
23:
/Users/rain/Project/async_simple/async_simple/util/ThreadPool.h/Users/rain/Project/async_simple/async_simple/util/ThreadPool.h::7575::10:10 : errorerror: : private field '_enableCoreBindings' is not used [-Werror,-Wunused-private-field]private field '_enableCoreBindings' is not used [-Werror,-Wunused-private-field]

    bool _enableCoreBindings;
    bool _enableCoreBindings;
         ^
         ^
1 error generated.
1 error generated.
make[2]: *** [async_simple/CMakeFiles/async_simple.dir/executors/SimpleExecutor.cpp.o] Error 1
make[2]: *** [async_simple/CMakeFiles/async_simple_static.dir/executors/SimpleExecutor.cpp.o] Error 1
make[1]: *** [async_simple/CMakeFiles/async_simple.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [async_simple/CMakeFiles/async_simple_static.dir/all] Error 2
In file included from /Users/rain/Project/async_simple/demo_example/block_echo_client.cpp:19:
In file included from /Users/rain/Project/async_simple/demo_example/asio_util.hpp:20:
In file included from /Users/rain/Project/async_simple/async_simple/executors/SimpleExecutor.h:23:
[...]

PR #110, opened 2022-05-18
Avoid unused member warning in ThreadPool

ChuanqiXu9:
## Why

Closing #109

### DIFF
--- a/async_simple/util/ThreadPool.h
+++ b/async_simple/util/ThreadPool.h
@@ -128,6 +128,10 @@ inline ThreadPool::ThreadPool(size_t threadNum, bool enableWorkSteal,
     std::vector<uint32_t> cpu_ids;
     if (_enableCoreBindings)
         getCurrentCpus(cpu_ids);
+#else
+    // Avoid unused member warning.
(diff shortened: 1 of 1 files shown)


async_simple/util/ThreadPool.h
```diff
@@ -128,6 +128,9 @@ inline ThreadPool::ThreadPool(size_t threadNum, bool enableWorkSteal,
     std::vector<uint32_t> cpu_ids;
     if (_enableCoreBindings)
         getCurrentCpus(cpu_ids);
+#else
+    // Avoid unused member warning.
+    (void)_enableCoreBindings;
```
RainMark, line 133: can we use `[[maybe_unused]]` attribute for member field?

ChuanqiXu9: Good idea! Done.

ChuanqiXu9: It would be ignored by GCC and cause another error. I would use the original style.

RainMark: ok

RainMark (approved):

PR #119, opened 2022-06-20
Decrease the scale of LazyPerf test under Debug mode

ChuanqiXu9:
## Why

The symmetric transfer is not part of C++ standard and it might be able to do at O0. So we chose to decrease the scale of LazyPerf further more.


RainMark (commented):
LGTM

RainMark (approved):

PR #122, opened 2022-06-27
Fix build failed on m1 mac clang-14

devillove084:


RainMark (approved):
LGTM

ChuanqiXu9 (approved):
LGTM. The reason may be you use new libcxx in Mac and we haven't tested that.

PR #128, opened 2022-08-08
allow temporary object as param in variadic collectAll

### LINKED ISSUE #127: The series function of collect* cause segment fault when param is a right reference to temporary object
### Search before asking

- [✔] I searched the [issues](https://github.com/alibaba/async_simple/issues) and found no similar issues.

The function collectAll/Any...  cause segment fault when param is a right reference to temporary object.

### Reproduction way
Just use temporary object as param in coroutine context.
```cpp
collectAll(Lazy<int>{},Lazy<float>{});
collectAll(std::vector<Lazy<int>>{});
//....
```

- [✔ ] Yes I am willing to submit a PR!

poor-circle:
## Why

Almostly Fix #127

## What is changing

In the variadic overload of collectAll, we can just copy the lazy<T> instead of reference. The copy is trivally.

However, it's expensive to copy in the vector<lazy<T>> overloaded. Maybe we can use span<lazy<T>> to instead, but it will cause interface change.

Luckily, only the empty temporary vector will cause crash(the non empty temporary vector will cause a compiler error in initialization)

## Example

see test in LazyTest.cpp

### DIFF
--- a/async_simple/coro/Collect.h
+++ b/async_simple/coro/Collect.h
@@ -350,8 +350,10 @@ inline auto collectAllPara(std::vector<LazyType<T>, IAlloc>&& input,
 // LazyType tasks in one thread,and producing a tuple of Try values containing
 // each of the results.
 template <template <typename> typename LazyType, typename... Ts>
-inline auto collectAll(LazyType<Ts>&&... inputs)
-    -> Lazy<std::tuple<Try<Ts>...>> {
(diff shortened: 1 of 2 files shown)


qicosmos (approved):
LGTM

ChuanqiXu9:
You need to sign the CLA license so that we could merge your pull request.

ChuanqiXu9 (approved):

PR #139, opened 2022-09-07
update CMakeLists.txt

PikachuHyA:
## Why

- to cooperate with cmake FetchContent better
- to handle macro definitions (e.g. `ASYNC_SIMPLE_HAS_NOT_AIO`) automatically
- use header-only 
- fix build fail when as subproject `ReadFiles`

```
target_link_libraries(xxx PRIVATE async_simple::async_simple_header_only)
```


ChuanqiXu9:
Let's try to do one thing in a PR. There are multiple goals in the PR and it may be confusing.

PR #140, opened 2022-09-07
fix build fail when as subproject

PikachuHyA:
## Why

sub pr of https://github.com/alibaba/async_simple/pull/139


ChuanqiXu9 (approved):
LGTM.

PR #141, opened 2022-09-13
update CMakeLists.txt: add header-only target

PikachuHyA:
sub pr of https://github.com/alibaba/async_simple/pull/139

## Why

async_simple can be used as header-only library if we do not use `SimpleExecutor` and uthread.

We can provide a header-only target of CMake, which is very convenient when async_simple is used as a subproject.

```
target_link_libraries(xxx PRIVATE async_simple::async_simple_header_only)
```

## What is changing

add `libasync_simple` target and it's alias  async_simple::async_simple_header_only


ChuanqiXu9 (commented):
It'd better to describe the intention clearly and directly.

PikachuHyA:
> It'd better to describe the intention clearly and directly.

description updated

ChuanqiXu9 (approved):
LGTM.

PR #142, opened 2022-09-13
fix: build fail with libc++

PikachuHyA:
## Why

`std::result_of` deprecated in C++17 and removed in C++20

see https://en.cppreference.com/w/cpp/types/result_of

## What is changing

replace `std:: result_of` by `std::invoke_result`


RainMark (approved):
LGTM

PR #157, opened 2022-09-18
Use std::variant insted of union in LazyPromisew

morebtcg:
## Why

#150


RainMark (commented):
lgtm

ChuanqiXu9 (approved):
LGTM. Would you like to refactor other similar patterns? (e.g., the FutureState)

PR #166, opened 2022-09-27
add benchmark: Lazy vs. Uthread

PikachuHyA:
## Why

add benchmark to compare Lazy and Thread

## What is changing

- add call depth benchmark
- add read file benchmark
- add switch benchmark

### DIFF
--- a/benchmarks/CallDepth.bench.cpp
+++ b/benchmarks/CallDepth.bench.cpp
@@ -27,148 +27,31 @@ using namespace async_simple::coro;
 using namespace async_simple::executors;
 using namespace async_simple::uthread;
 namespace fs = std::filesystem;
-static int task_num = 10;
-struct Service {
-    void run() { service1(); }
+static int core_num = 1;
 
-    void service1() {
-        n--;
-        if (n == 0) {
-            read_db();
-            return;
-        }
-        int call_idx = n % 2;
-        if (call_idx == 0) {
-            return service2();
-        } else {
-            return service3();
-        }
+int sum(int n) {
+    if (n == 0) {
+        return 0;
     }
+    return n + sum(n - 1);
+}
 
-    void service2() {
-        n--;
-        if (n == 0) {
-            read_db();
-            return;
-        }
-        int call_idx = n % 2;
-        if (call_idx == 0) {
-            service1();
-        } else {
-            return service3();
-        }
-    }
-
-    void service3() {
-        n--;
-        if (n == 0) {
-            read_db();
-            return;
-        }
-        int call_idx = n % 2;
-        if (call_idx == 0) {
-            return service1();
-        } else {
-            return service2();
-        }
-    }
-
-    void read_db() {
-        using namespace async_simple::uthread;
-        std::vector<std::function<void()>> task_list;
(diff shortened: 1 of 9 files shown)


benchmarks/CallDepth.bench.cpp
```diff
@@ -0,0 +1,204 @@
...
+        state.PauseTiming();
+        std::vector<std::function<void()>> task_list;
+        task_list.reserve(task_num);
+        std::atomic<bool> running = true;
+        int m = 1;
+        for (int i = 0; i < task_num; ++i) {
+            task_list.emplace_back([n, m, s, e = &e]() {
+                Service service{
+                    .n = n, .m = m, .filename_ = s.c_str(), .e_ = e};
+                service.run();
```
RainMark, line 164: coroutine call depth perf test should not include the io path?

benchmarks/ReadFileUtil.hpp
```diff
@@ -0,0 +1,132 @@
...
+    task_list.reserve(num);
+    for (int i = 0; i < num; ++i) {
+        task_list.template emplace_back(
+            [&s, e]() { Uthread_read_file(s.c_str(), e); });
+    }
+    collectAll<Launch::Schedule>(task_list.begin(), task_list.end(), e);
+}
+
+template <typename FileDescriptor, typename Buffer, typename Executor>
+struct ReadSomeAwaiter {
```
RainMark, line 65: try to use FutureAwaiter to instead of self-defined awaiter?

benchmarks/CallDepth.bench.cpp
```diff
@@ -0,0 +1,204 @@
...
+        auto f = [&s, n, e = &e]() mutable -> Lazy<void> {
+            std::vector<Lazy<uint64_t>> lazy_list;
+            lazy_list.reserve(task_num);
+            int m = 1;
+            for (int i = 0; i < task_num; ++i) {
+                lazy_list.push_back(
+                    AsyncService().run(s.c_str(), n, e->getIOExecutor(), m));
+            }
+            [[maybe_unused]] auto ret =
+                co_await collectAllPara(std::move(lazy_list));
```
RainMark, line 198: also should not include task schedule latency, collectAllPara will occurred the task schedule.

benchmarks/CallDepth.bench.cpp
```diff
@@ -0,0 +1,204 @@
...
+using namespace async_simple;
+using namespace async_simple::coro;
+using namespace async_simple::executors;
+using namespace async_simple::uthread;
+namespace fs = std::filesystem;
+static int task_num = 10;
+struct Service {
+    void run() { service1(); }
+
+    void service1() {
```
RainMark, line 34: we already has call depth test case in https://github.com/alibaba/async_simple/blob/main/async_simple/coro/test/LazyTest.cpp#L1307, maybe you can move to benchmark module. And add the other perf test like coroutine switch.

benchmarks/ReadFileUtil.hpp
PikachuHyA: done

ChuanqiXu9 (approved):
LGTM.

PR #247, opened 2022-11-23
Adjust CI: add gcc-11 and gcc-12

4kangjc:
## Why
https://github.com/alibaba/async_simple/issues/234


ChuanqiXu9 (approved):
LGTM. Thanks!

PR #283, opened 2023-01-14
Fix: Deducing This is not yet implemented in gcc and clang

4kangjc:
## Why
我的疏忽，目前`Deducing This`这个特性在gcc和clang上还没实现，所以编译选项改成`-std=c++23`会编译错误


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #286, opened 2023-02-08
private to protected

### LINKED ISSUE #285: can not override a private virtual function in Executor
### Search before asking

```virtual void schedule(Func func, Duration dur)``` [the function](https://github.com/alibaba/async_simple/blob/main/async_simple/Executor.h#L125) is private virtual function, can't override it, we should change it to protected and allow user override it.

- [x] Yes I am willing to submit a PR!

qicosmos:
## Why

Closes #285

### DIFF
--- a/async_simple/Executor.h
+++ b/async_simple/Executor.h
@@ -121,7 +121,7 @@ public:
         throw std::logic_error("Not implemented");
     }
 
-private:
+protected:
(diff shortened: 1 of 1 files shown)


ChuanqiXu9 (approved):
LGTM. Thanks.

Issue #302, opened 2023-04-13
async_simple/coro/Lazy.h:239:16: error: moving a temporary object prevents copy elision [-Werror=pessimizing-move]


topazus:
### Reproduction way
```bash
❯ git clone --depth=1 https://github.com/alibaba/async_simple
Cloning into 'async_simple'...
remote: Enumerating objects: 959, done.
remote: Counting objects: 100% (959/959), done.
remote: Compressing objects: 100% (638/638), done.
remote: Total 959 (delta 345), reused 810 (delta 311), pack-reused 0
Receiving objects: 100% (959/959), 2.50 MiB | 4.30 MiB/s, done.
Resolving deltas: 100% (345/345), done.

❯ git log
commit 5439d81bfd7fe83baf8b2291fdd197414b7cc43a (grafted, HEAD -> main, origin/main, origin/HEAD)
Author: kang jinci <1158035520@qq.com>
Date:   Thu Apr 13 10:10:21 2023 +0800

    fix: GCC 12 issues a spurious uninitialized-var warning (#301)

ruby in 🌐 fedora in ~/build took 2s 
❯ cd async_simple/

ruby in 🌐 fedora in async_simple on  main is 📦 v1.0.0 via △ v3.26.2 via 🌙 v5.4.4 via  v19.8.1 via 🐍 v3.11.2 
❯ cmake . -B build 
-- The CXX compiler identification is GNU 13.0.1
-- The C compiler identification is GNU 13.0.1
-- The ASM compiler identification is GNU
-- Found assembler: /usr/lib64/ccache/cc
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib64/ccache/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- CMAKE_MODULE_PATH: /home/ruby/build/async_simple/cmake
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- FindAio.cmake: 
-- Found Benchmark: /usr/lib64/libbenchmark.so  
-- aio found
[...]

PR #303, opened 2023-04-13
fix: GCC 13 compilation failed

4kangjc:
## Why
Closes https://github.com/alibaba/async_simple/issues/302


async_simple/FutureState.h
```diff
@@ -195,7 +195,7 @@ class FutureState {
     //  ONLY_RESULT: promise.setValue called
     //  ONLY_CONTINUATION: future.thenImpl called
     void setResult(Try<T>&& value) {
-#if __GNUC__ != 12
+#if __GNUC__ < 12
```
ChuanqiXu9, line 198:
```suggestion
#if !defined(__GNUC__) || __GNUC__ < 12
```

PR #313, opened 2023-06-01
refactor: remove unnecessary lambda wrap

chloro-pn:
## What is changing

Remove some unnecessary lambda wrap as std::coroutine_handle<> is callable object


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #329, opened 2023-06-17
Remove SimpleIOExecutor and SimpleExecutor from libasync_simple

4kangjc:
SimpleIOExecutor and IOExecutor are for testing and examples only, not part of libasync_simple


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #330, opened 2023-06-17
fix: benchmark::DoNotOptimize(const Tp&) is deprecated

4kangjc:
The const-ref version of this method can permit undesired compiler optimizations in benchmarks


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #338, opened 2023-07-26
Introduce get_return_object_on_allocation_failure to Lazy

ChuanqiXu9:
This patch tries to introduce `get_return_object_on_allocation_failure()` to the promise_type of Lazy.

Why do we want to introduce `get_return_object_on_allocation_failure()`? Since
a coroutine will be roughly converted to:

```C++
void *frame_addr = ::operator new(required size);
__promise_ = new (frame_addr) __promise_type(...);
__return_object_ = __promise_.get_return_object();
co_await __promise_.initial_suspend();
try {
    function-body
} catch (...) {
    __promise_.unhandled_exception();
}
co_await __promise_.final_suspend();
```

Then we can find that the coroutine should be nounwind (noexcept) naturally if
the constructor of the promise_type, the get_return_object() function, the
initial_suspend, the unhandled_exception(), the final_suspend and the allocation
function is noexcept.

For the specific coroutine type, Lazy, all the above except the allocation function
is noexcept. So that we can make every Lazy function noexcept naturally if we make
the allocation function nothrow. This is the reason why we want to introduce
`get_return_object_on_allocation_failure()` to Lazy.

Note that the optimization may not work in some platforms due the ABI limitations.
Since they need to consider the case that the destructor of an exception can throw
exceptions.

### DIFF
--- a/async_simple/coro/Lazy.h
+++ b/async_simple/coro/Lazy.h
@@ -145,7 +145,11 @@ public:
     LazyPromise() noexcept {}
     ~LazyPromise() noexcept {}
 
+    LazyPromise(std::exception_ptr ptr) { _value = ptr; }
+
     Lazy<T> get_return_object() noexcept;
+
+    static Lazy<T> get_return_object_on_allocation_failure() noexcept;
 
     template <typename V>
     void return_value(V&& value) noexcept(
@@ -190,7 +194,13 @@ public:
 template <>
 class LazyPromise<void> : public LazyPromiseBase {
 public:
+    LazyPromise() noexcept {}
+    ~LazyPromise() noexcept {}
+    LazyPromise(std::exception_ptr ptr) { _exception = ptr; }
+
     Lazy<void> get_return_object() noexcept;
+    static Lazy<void> get_return_object_on_allocation_failure() noexcept;
+
     void return_void() noexcept {}
     void unhandled_exception() noexcept {
         _exception = std::current_exception();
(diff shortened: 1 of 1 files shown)


async_simple/coro/Lazy.h
```diff
@@ -551,6 +561,115 @@ inline Lazy<void> detail::LazyPromise<void>::get_return_object() noexcept {
...
+        static void __resume_func(
+            __allocation_failed_coroutine_frame_ty* frame) {
+            if (frame->__promise_._continuation)
+                frame->__promise_._continuation.resume();
+        }
+
+        void (*__resume_)(__allocation_failed_coroutine_frame_ty*) =
+            __resume_func;
+        void (*__destroy_)(void*) = __dummy_destroy_func;
+        class LazyPromise<T> __promise_;
```
RainMark, line 637: 这里的exception什么时候被用户touch到

async_simple/coro/Lazy.h
```diff
@@ -551,6 +561,115 @@ inline Lazy<void> detail::LazyPromise<void>::get_return_object() noexcept {
...
+                __promise_._exception =
+                    std::make_exception_ptr(std::bad_alloc{});
+            } else {
+                __promise_._value = std::make_exception_ptr(std::bad_alloc{});
+            }
+        }
+    };
+
+    static inline __allocation_failed_coroutine_frame_ty
+        __allocation_failed_coroutine_frame;
```
RainMark, line 650: 多个线程同时出现了bad alloc，是否会有race

ChuanqiXu9: hmmm 好像是不太对，这种情况因为平时不会耗尽内存所以没遇到，我先 revert 吧

RainMark: 其实这里有点尴尬，bad alloc时一般都是内存比较紧张了。只好提前分配比较合理，thread local是不是可以解，没特别想清楚

PR #342, opened 2023-09-18
fix error message for Try.h

StephenRi:
async-simple Try.h中存在报错信息拼写错误
old："Try object do not has on error"
fixed: "Try object do not has an error"

#343


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #360, opened 2024-01-08
fix: update CMakeLists for AppleClang 15.x Support

faker2048:
## Why

The current CMake settings do not specifically address AppleClang 15.x versions, which is needed to handle certain compiler errors.

macos 14.2.1 (m2) version of clang:
```
message("Compiler ID: ${CMAKE_CXX_COMPILER_ID}")
message("Compiler Version: ${CMAKE_CXX_COMPILER_VERSION}") 

# Compiler ID: AppleClang
# Compiler Version: 15.0.0.15000040
```

## What is changing

Added a condition in CMakeLists.txt to apply the -Wno-unsequenced option specifically for AppleClang 15.x


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #364, opened 2024-01-25
make lsp happy

yinghaoyu:
1. For `.clangd`, see also https://github.com/clangd/clangd/issues/662#issuecomment-1447233948.
2. `async_simple/Future.h` and `async_simple/Promise.h` include recursively.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,3 @@
+CompileFlags:
+  Add: [-std=c++20, -Wno-unknown-warning-option]
+  Remove: [-m*, -f*]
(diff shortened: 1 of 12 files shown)


ChuanqiXu9 (approved):
LGTM.  Thanks.

Let's see if the CI feels good about it.

ChuanqiXu9:
It looks like it is a breaking change to require the users to include `Promise.h` explicitly. This may be not good. Can we teach clangd to not complain about it?

Also the formatter looks not happy too.

yinghaoyu:
`std::promise` as a friend class for `std::future`, both of them are in the same file. Emmmm, maybe it's a `breaking change`. Please ignore this commit.

yinghaoyu:
My clang is out of date, so the formatter is not fit.

PR #366, opened 2024-01-30
fix find_package() & add option to disable asan

poor-circle:
## What is changing

1. the target should add compile options `-fcoroutines`, so that user don't need to add it manually.
2. add option to disable asan.


async_simple/CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ if(UTHREAD)
 endif()
 
 # If there is no Uthread, async_simple is a header only library 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  target_compile_options(async_simple_header_only INTERFACE "-fcoroutines")
```
ChuanqiXu9, line 27: GCC enabled coroutines by default after 11. We need to check the version here. A message may be helpful here too.

async_simple/CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ if(UTHREAD)
   list(APPEND SRCS ${uthread_asm_src})
 endif()
 
+
+if (NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") AND (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
```
ChuanqiXu9, line 26:
```suggestion
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```

async_simple/CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ if(UTHREAD)
...
+  if (NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.0.0" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "10.0.0")
```
ChuanqiXu9, line 27:
```suggestion
  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.0.0")
```

async_simple/CMakeLists.txt
poor-circle: add check for gcc and gcc version

poor-circle:
We need release a new version 1.3, then i will add async_simple to vcpkg.

ChuanqiXu9 (approved):
LGTM. Thanks.

PR #369, opened 2024-02-02
[doc] add description for install by vcpkg, simply install document

poor-circle:
## Why

simply document to helper newcomer install and user async-simple.


ChuanqiXu9 (approved):
LGTM. Thanks.

Generally I don't like `[no ci]` since we don't have CI pressure.  But this is nit.

PR #370, opened 2024-03-07
[fix] fix compile

qicosmos:
## Why

fix compile error on gcc10.2 or lower version;

fix compile error on msvc;

## What is changing

add no-unused-value when gcc10.2 or lower;

use (std::min) instead of std::min to avoid conflict on windows;

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,6 @@ else()
         -Wno-deprecated-register
         -Wno-mismatched-new-delete
         -Wno-deprecated-declarations
-        -Wno-unused-value
         -D_FILE_OFFSET_BITS=64
         -fPIC
         -Wall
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ else()
         -Wno-deprecated-register
         -Wno-mismatched-new-delete
         -Wno-deprecated-declarations
+        -Wno-unused-value
```
ChuanqiXu9, line 66: Generally it is bad to have unused values, what is the warning? Can we try to fix that?

async_simple/coro/Collect.h
```diff
@@ -371,7 +371,7 @@ inline auto collectAllWindowedImpl(size_t maxConcurrency,
     }
     size_t start = 0;
     while (start < input_size) {
-        size_t end = std::min(input_size, start + maxConcurrency);
+        size_t end = (std::min)(input_size, start + maxConcurrency);
```
ChuanqiXu9, line 374: Let's add a comment to explain this.

CMakeLists.txt
qicosmos:
```
[build] [ 85%] Building CXX object demo_example/CMakeFiles/CountChar.dir/CountChar.cpp.o
[build] async_simple/demo_example/CountChar.cpp: In function ‘async_simple::coro::Lazy<int> CountTextChar(const Texts&, char)’:
[build] async_simple/demo_example/CountChar.cpp:50:1: error: statement has no effect [-Werror=unused-value]
[build]    50 | }
[build]       | ^
[build] At global scope:
```

```c++
Lazy<int> CountTextChar(const Texts &Content, char c) {
    int Result = 0;
    for (const auto &line : Content)
        Result += co_await CountLineChar(line, c);
    co_return Result;
}
```

I think it's a bug of gcc10.

ChuanqiXu9: Agreed. But disable `-Wno-unused-value` looks over acted. I feel we can add a `(void)Result;` before `co_return` to eliminate the warning.

PR #371, opened 2024-03-08
Extend collect any

qicosmos:
## Why

extend collectAny with callback, similar with go/kotlin select coroutine function.

## What is changing

add two new collectAny methods

## Example

collectAny with variadic callback functions:
```c++
    auto test0 = []() -> Lazy<Unit> { co_return Unit{}; };
    auto test1 = []() -> Lazy<int> { co_return 42; };
    auto test2 = [](int val) -> Lazy<std::string> {
        co_return std::to_string(val);
    };

    int call_count = 0;
    index =
        syncAwait(collectAny(std::pair{test0(), [&](auto) { call_count++; }},
                             std::pair{test1(),
                                       [&](auto val) {
                                           call_count++;
                                           EXPECT_EQ(val.value(), 42);
                                       }},
                             std::pair{test2(42), [&](auto val) {
                                           call_count++;
                                           EXPECT_EQ("42", val.value());
                                       }}));
    EXPECT_EQ(1, call_count);
```

collectAny with vector callback function:
```c++
    auto test0 = []() -> Lazy<int> { co_return 41; };
    auto test1 = []() -> Lazy<int> { co_return 42; };

    std::vector<Lazy<int>> input;
    input.push_back(test0());
    input.push_back(test1());

    syncAwait(collectAny(std::move(input), [](size_t index, Try<int> val) {
        if (index == 0) {
            EXPECT_EQ(val.value(), 41);
        } else {
            EXPECT_EQ(val.value(), 42);
        }
    }));
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -99,7 +99,3 @@ DisableFormat: true
 ---
 Language: Json
 DisableFormat: true
----
-Language: ObjC
-DisableFormat: true
-
--- a/.clang-format
+++ b/.clang-format
@@ -99,3 +99,7 @@ DisableFormat: true
 ---
 Language: Json
 DisableFormat: true
+---
+Language: ObjC
+DisableFormat: true
+
--- a/.github/workflows/conan.yml
+++ b/.github/workflows/conan.yml
@@ -22,11 +22,7 @@ jobs:
 
       - name: deps
         run: |
-          sudo apt install libaio-dev libgtest-dev -y
-          cd /usr/src/gtest
-          sudo mkdir build && cd build
-          sudo cmake .. && sudo make
-          sudo apt install libgmock-dev -y
+          sudo apt install libaio-dev
 
       - name: Install newer Clang
         run: |
--- a/.github/workflows/conan.yml
+++ b/.github/workflows/conan.yml
@@ -22,7 +22,11 @@ jobs:
 
       - name: deps
         run: |
-          sudo apt install libaio-dev
+          sudo apt install libaio-dev libgtest-dev -y
+          cd /usr/src/gtest
+          sudo mkdir build && cd build
+          sudo cmake .. && sudo make
+          sudo apt install libgmock-dev -y
 
       - name: Install newer Clang
         run: |
--- a/async_simple/coro/Collect.h
+++ b/async_simple/coro/Collect.h
@@ -506,46 +506,6 @@ inline auto CollectAnyVariadicImpl(LazyType<Ts>&&... inputs) {
                            CollectAnyVariadicAwaiter<LazyType, Ts...>>;
     return AT(std::move(inputs)...);
 }
-
-// collectAny with std::pair<Lazy, CallbackFunction>
-template <typename Tuple>
-class Helper {
-public:
-    Helper(Tuple tuple) : _callback_tuple(std::move(tuple)) {}
-
-    async_simple::coro::Lazy<size_t> operator()(auto&&... lazy) {
-        auto result = co_await collectAny(std::move(lazy)...);
-
-        tupleSwitch(result.index(), _callback_tuple, result);
-        co_return result.index();
-    }
-
-private:
-
-    template <class Result, std::size_t... Is>
-    void tupleSwitchImpl(std::size_t i, Tuple& t, Result& result,
-                           std::index_sequence<Is...>) {
-        ((void)(i == Is && (std::get<Is>(t)(std::move(std::get<Is>(result))), false)),
-         ...);
-    }
-
-    template <class Result>
-    void tupleSwitch(std::size_t i, Tuple& t, Result& result) {
-        tupleSwitchImpl(i, t, result,
-                          std::make_index_sequence<std::tuple_size_v<Tuple>>{});
-    }
-
-    Tuple _callback_tuple;
-};
-
-template <std::size_t N>
-inline constexpr auto splitPairTuple(auto&& t) noexcept {
-    return [&t]<auto... I>(std::index_sequence<I...>) {
-        return std::make_pair(
-            std::make_tuple(std::move(std::get<I>(t).first)...),
-            std::make_tuple(std::move(std::get<I>(t).second)...));
-    }(std::make_index_sequence<N>());
-}
 }  // namespace detail
 
 template <typename T, template <typename> typename LazyType,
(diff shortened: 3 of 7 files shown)


async_simple/coro/Collect.h
```diff
@@ -507,6 +507,45 @@ inline auto CollectAnyVariadicImpl(LazyType<Ts>&&... inputs) {
...
+template <typename Tuple>
+class Helper {
+public:
+    Helper(Tuple tuple) : _callback_tuple(std::move(tuple)) {}
+
+    async_simple::coro::Lazy<size_t> operator()(auto&&... lazy) {
+        auto result = co_await collectAny(std::move(lazy)...);
+
+        tupleSwitch(result.index(), _callback_tuple, result);
+        co_return result.index();
```
ChuanqiXu9, line 520: Let's try to avoid adding layers in implementing fundamental functions by new co_await. It may results new allocations/deallocations, which may be a performance concern.

async_simple/coro/test/LazyTest.cpp
```diff
@@ -1104,6 +1104,104 @@ TEST_F(LazyTest, testCollectAllVariadic) {
...
+    int test_value = 0;
+    index = syncAwait(collectAny(std::pair{test3().via(&e1),
+                                           [&](auto val) {
+                                               test_value = val.value();
+                                               EXPECT_EQ(42, test_value);
+                                           }},
+                                 std::pair{test4(41).via(&e1), [&](auto val) {
+                                               test_value = val.value();
+                                               EXPECT_EQ(41, test_value);
+                                           }}));
```
ChuanqiXu9, line 1150: It will be better if we can avoid writing `std::pair` and try to use `{, }`. It may be possible to construct a pair from the initializer list.

qicosmos: I think maybe it's impossible: https://stackoverflow.com/questions/65071836/template-argument-deduction-for-implicit-pair

can't use std::initializer_list to avoid std::pair with brace initialize.

another reason, can't use std::initializer_list because it requires types are the same, not variadic template.

async_simple/coro/Collect.h
```diff
@@ -521,6 +520,35 @@ inline auto collectAny(LazyType<Ts>... awaitables) {
...
+template <typename... Ts>
+inline Lazy<size_t> collectAny(Ts... args) {
+    auto tuple = std::make_tuple(std::move(args)...);
+    auto lazy_pair = [&tuple]<size_t... I>(std::index_sequence<I...>) {
+        return std::make_pair(
+            collectAny(std::move(std::get<0>(std::get<I>(tuple)))...),
+            std::make_tuple(std::move(std::get<1>(std::get<I>(tuple)))...));
+    }(std::make_index_sequence<sizeof...(Ts)>());
+
+    auto result = co_await std::move(lazy_pair.first);
```
ChuanqiXu9, line 533: This still introduces new coroutines, which may be a performance concern. Let's try to avoid the new creation of coroutines as much as possible.

async_simple/coro/test/LazyTest.cpp
ChuanqiXu9: Got it.

ChuanqiXu9 (commented):
Also we need to add documents for this.

async_simple/coro/Collect.h
```diff
@@ -123,25 +127,120 @@ struct CollectAnyAwaiter {
...
+        }
     }
 
     std::vector<LazyType, InAlloc> _input;
     std::shared_ptr<ResultType> _result;
+    std::tuple<Function...> _callback;
+};
+
+template <typename... Ts>
+struct CollectAnyVariadicPairAwaiter {
```
ChuanqiXu9, line 161: Can we try to reuse this with `CollectAnyVariadicAwaiter`?

async_simple/coro/Collect.h
```diff
@@ -123,25 +127,120 @@ struct CollectAnyAwaiter {
...
+                        });
+                }(),
+                ...);
+        }
+        (std::make_index_sequence<sizeof...(Ts)>());
+    }
+
+    auto await_resume() {
+        assert(_result != nullptr);
+        return std::move(_result->value());
```
ChuanqiXu9, line 226: Why we don't return value with callbacks for vectors but returning here?

ChuanqiXu9 (commented):
We also need to update the document for such new features.

async_simple/coro/Collect.h
```diff
@@ -507,12 +614,22 @@ inline auto CollectAnyVariadicImpl(LazyType<Ts>&&... inputs) {
...
+}
 }  // namespace detail
 
 template <typename T, template <typename> typename LazyType,
-          typename IAlloc = std::allocator<LazyType<T>>>
-inline auto collectAny(std::vector<LazyType<T>, IAlloc>&& input) {
-    return detail::collectAnyImpl(std::move(input));
+          typename IAlloc = std::allocator<LazyType<T>>, typename... Function>
+inline auto collectAny(std::vector<LazyType<T>, IAlloc>&& input,
+                       Function... func) {
```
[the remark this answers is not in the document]
qicosmos: for reuse code, user don't pass a callback function, the code go back to original version, otherwise go to callback version

async_simple/coro/Collect.h
qicosmos: The code will become ugly, the input type and return type are different, a new awaiter mabye more clean and clear?

async_simple/coro/Collect.h
qicosmos: variadic version no need index, because the sequence is the index. the index for vector version.

async_simple/coro/Collect.h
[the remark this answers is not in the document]
ChuanqiXu9: I feel better with offering an overload here to make things explciitly.

async_simple/coro/Collect.h
ChuanqiXu9: Fine.

async_simple/coro/Collect.h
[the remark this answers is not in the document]
qicosmos: OK, remove the variadic template, use a template parameter: Callback, the default Callback type is Unit, and  the 
default Callback member variable can be optimized by compiler via [[no_unique_address]].

.clang-format
```diff
@@ -99,3 +99,7 @@ DisableFormat: true
 ---
 Language: Json
 DisableFormat: true
+---
+Language: ObjC
+DisableFormat: true
```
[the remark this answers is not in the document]
ChuanqiXu9: Please remove the unrelated change.

async_simple/coro/test/LazyTest.cpp
```diff
@@ -916,192 +916,305 @@ TEST_F(LazyTest, testCollectAllWithAllocator) {
...
-        auto v_try_int = std::get<0>(std::move(out_tuple));
-        auto v_try_double = std::get<1>(std::move(out_tuple));
-        auto v_try_string = std::get<2>(std::move(out_tuple));
-
-        EXPECT_EQ(2, v_try_int.value());
-        EXPECT_DOUBLE_EQ(2.2, v_try_double.value());
-        EXPECT_STREQ("testCollectAllVariadic", v_try_string.value().c_str());
-
-        CHECK_EXECUTOR(&e1);
+// TEST_F(LazyTest, testCollectAllVariadic) {
```
ChuanqiXu9, line 919: Why the tests got commented out?

async_simple/coro/test/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 file(GLOB coro_test_src "*.cpp")
+add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/bigobj>")
```
ChuanqiXu9, line 2: Why do we need this?

ChuanqiXu9 (commented):
We need to document this

async_simple/coro/test/LazyTest.cpp
qicosmos: just for test, will revert.

async_simple/coro/test/CMakeLists.txt
qicosmos: windows ci failed before,
https://github.com/alibaba/async_simple/actions/runs/8200305183/job/22426819820#step:4:304

.clang-format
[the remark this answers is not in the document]
qicosmos: to solve this error:https://github.com/alibaba/async_simple/actions/runs/8200305198/job/22426819367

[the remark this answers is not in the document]
ChuanqiXu9: This must not be correct. There is nothing related to ObjC.

.github/workflows/conan.yml
```diff
@@ -27,6 +27,13 @@ jobs:
           sudo mkdir build && cd build
           sudo cmake .. && sudo make
           sudo apt install libgmock-dev -y
+
+      - name: Install newer Clang
+        run: |
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x ./llvm.sh
+          sudo ./llvm.sh 17
```
ChuanqiXu9, line 35: We should do one thing in one patch. Please move this to other PR.

qicosmos: OK, when test passed will create a new pr to upgrade clang for conan.

async_simple/coro/test/CMakeLists.txt
ChuanqiXu9: This shouldn't be related to this patch. Please remove the unrelated change.

qicosmos: OK, will create a new PR to add this compile option for msvc.

qicosmos: done in this PR: https://github.com/alibaba/async_simple/pull/375

PR #385, opened 2024-07-24
coro/Generator: correct Generator alias of std::generator 

tchaikov:
before this change, the build fails, because `AlloAllocator` is not defined when the standard library provides `std::generator`, and the tree is built with C++23.

in this change, we replace `AlloAllocator` with `Allocator`. and the tree builds fine.

also, we include `<generator>` only if `__cpp_lib_generator` is defined. as the existence of the header does not imply that we are building with C++23.


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #399, opened 2024-11-05
Fix collectAllPara dont use SimpleCollectAllVariadicAwaiter

poor-circle:
## Why

it seems that After #277 , collectAllPara dont use `SimpleCollectAllVariadicAwaiter`, which is faster by provide `coAwait` method

## What is changing

Fix code that collectAllPara will call `SimpleCollectAllVariadicAwaiter`.


ChuanqiXu9 (approved):
LGTM

PR #402, opened 2024-11-15
Add cancellation support

poor-circle:
## Why

See document

## What is changing

Breakchanges:

1. Components such as Yield and `spinlock` may throw exceptions.
2. The return values of `CollectAll`/`CollectAny` have been changed to lazy, and the callback versions of `collectAll`/`collectAny` have been removed.


poor-circle:
@ChuanqiXu9 It seems that mac ci & bazel ci are broken.

ChuanqiXu9:
> @ChuanqiXu9 It seems that mac ci & bazel ci are broken.

CC @PikachuHyA

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+            auto newManager = new Signal::HandlerManager(index, handlerManager);
+            // It's ok here, write only has once thread
+            _node->_handlerManager.store(newManager, std::memory_order_release);
+            _handlerTables.emplace(index, &newManager->_handler);
+            return &newManager->_handler;
+        }
+    }
+
+    // TODO: may be flatmap better?
+    std::unordered_map<uint8_t, std::atomic<Signal::Handler*>*> _handlerTables;
```
ChuanqiXu9, line 317: Why atomic? We should have banned manipulating the same slot simultaneously.

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+                _next = _next->_next;
+                delete tmp;
+            }
+            delete this;
+        }
+        ~HandlerManager();
+        static inline Handler emittedTag;
+    };
+    struct PrivateConstructTag {};
+    struct Node {
```
ChuanqiXu9, line 77: `Node` is actually a shared state between `Slot` and `Signal`. It is not good to make it a subclass of `Signal`. Let's move it and renaming to `SignalSlotSharedState`. The list of `SignalSlotSharedState` can be another data structure.

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+    all = UINT64_MAX,
+};
+
+class Slot;
+
+class Signal : public std::enable_shared_from_this<Signal> {
+private:
+    using Handler = util::move_only_function<void(SignalType, Signal*)>;
+    using TransferHandler = util::move_only_function<void(SignalType)>;
+    struct HandlerManager {
```
ChuanqiXu9, line 60: `HandlerManager` should be a subclass of `Node`

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+            _filter.store(filter, std::memory_order_release);
+        }
+        SignalType getFilter() const noexcept {
+            return _filter.load(std::memory_order_acquire);
+        }
+        void invoke(SignalType type, SignalType signal, Signal* self,
+                    std::atomic<Handler*>& handlerPtr);
+        void operator()(SignalType type, Signal* self);
+        std::atomic<HandlerManager*> _handlerManager;
+        std::atomic<TransferHandler*> _transferHandler;
```
ChuanqiXu9, line 104: Is it really necessary to make it atomic?

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+    // after signal triggered will always success and return true.
+    all = UINT64_MAX,
+};
+
+class Slot;
+
+class Signal : public std::enable_shared_from_this<Signal> {
+private:
+    using Handler = util::move_only_function<void(SignalType, Signal*)>;
+    using TransferHandler = util::move_only_function<void(SignalType)>;
```
ChuanqiXu9, line 59: Move this to `Node` too.

async_simple/Signal.h
ChuanqiXu9, line 104: `TransferHandler` is not a good name. Rename it to `ChainedSignalHandler`

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+        }
+        SignalType getFilter() const noexcept {
+            return _filter.load(std::memory_order_acquire);
+        }
+        void invoke(SignalType type, SignalType signal, Signal* self,
+                    std::atomic<Handler*>& handlerPtr);
+        void operator()(SignalType type, Signal* self);
+        std::atomic<HandlerManager*> _handlerManager;
+        std::atomic<TransferHandler*> _transferHandler;
+        std::atomic<SignalType> _filter;
```
ChuanqiXu9, line 105: Is it necessary to make `_filter` atomic? Since it is only accessed by a `Slot`, which shouldn't be accessed simultaneously

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+        Node* _next;
+    };
+    void registSlot(Slot* slot, SignalType filter);
+
+    inline static bool isMultiTriggerSignal(SignalType type) {
```
ChuanqiXu9, line 110: Why "must"?

ChuanqiXu9, line 110:
```suggestion
   static bool isMultiTriggerSignal(SignalType type) {
```

Also it looks better to not be a member function.

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+        return type >= (1ull << 32);
+    }
+
+    inline SignalType UpdateState(std::atomic<uint64_t>& state,
```
ChuanqiXu9, line 114:
```suggestion
   SignalType UpdateState(std::atomic<uint64_t>& state,
```

async_simple/Signal.h
poor-circle: signal->emit() may modify handler when slot emplace()/clear() handler concurrently, so it should be atomic.

async_simple/Signal.h
poor-circle: Yes. The thread which call signal->emit() and the thread which call slot->emplace() will read/write the node concurrently. So it should be atomic.

async_simple/Signal.h
poor-circle: When we call signal->emit() will also read _filter to filter the signal. So it should be atomic.

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+        auto handler =
+            std::make_unique<Signal::Handler>(std::forward<Args>(args)...);
+        logicAssert(static_cast<bool>(*handler),
+                    "CancellationSlot dont allow emplace empty handler");
+        auto oldHandlerPtr = loadHandler<true>(type);
+        auto oldHandler = oldHandlerPtr->load(std::memory_order_acquire);
+        if (oldHandler == &Signal::HandlerManager::emittedTag) {
+            return false;
+        }
+        auto new_handler = handler.release();
```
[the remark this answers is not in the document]
poor-circle: in line 184 we dont need delete it manully

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+        uint8_t index = std::countr_zero(static_cast<uint64_t>(type));
+        if (auto iter = _handlerTables.find(index);
+            iter != _handlerTables.end()) {
+            return iter->second;
+        }
+        if constexpr (!should_emplace) {
+            return nullptr;
+        } else {
+            auto handlerManager =
+                _node->_handlerManager.load(std::memory_order_acquire);
```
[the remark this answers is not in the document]
poor-circle: _node never be nullptr if slot constructed, it's init by constructors.

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+    // already been triggered(and signal can't trigger twice), or the slot is
+    // disconnected.
+    template <typename... Args>
+    [[nodiscard]] bool emplace(SignalType type, Args&&... args) {
+        static_assert(sizeof...(args) >= 1,
+                      "we dont allow emplace an empty signal handler");
+        logicAssert(std::popcount(static_cast<uint64_t>(type)) == 1,
+                    "It's not allow to emplace for multiple signals");
+        // signal has triggered and cant trigger twice
+        if (type < (1ull << 32) && (signal()->state() & type)) {
```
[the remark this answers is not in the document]
poor-circle: _node never be nullptr if slot constructed, it's init by constructors.

async_simple/Signal.h
```diff
@@ -0,0 +1,453 @@
...
+#include <unordered_map>
+#include <utility>
+
+#include "async_simple/Common.h"
+#include "util/move_only_function.h"
+#endif  // ASYNC_SIMPLE_USE_MODULES
+
+namespace async_simple {
+
+enum SignalType : uint64_t {
```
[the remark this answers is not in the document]
poor-circle: User may extend this enum, so enum maybe better than enum class

poor-circle:
@ChuanqiXu9 All conversation resolved.

async_simple/coro/test/LazyTest.cpp
```diff
@@ -1439,6 +1353,245 @@ TEST_F(LazyTest, testCollectAnyVariadic) {
...
+    mylocal(std::string sv) : LazyLocalBase(&tag), name(std::move(sv)) {}
+    std::string& hello() { return name; }
+    std::string name;
+    static bool classof(const LazyLocalBase* base) {
+        return base->getTypeTag() == &tag;
+    }
+    inline static char tag;
+};
+
+auto my_sleep_impl = [](std::chrono::milliseconds ms, SignalType expected_type,
```
ChuanqiXu9, line 1366: Please use function instead of global lambda

async_simple/coro/test/LazyTest.cpp
```diff
@@ -1111,120 +1139,6 @@ TEST_F(LazyTest, testCollectAllVariadic) {
     syncAwait(testCollectAllPara().via(&e1));
 }
 
-TEST_F(LazyTest, testCollectAnyWithCallbackVariadic) {
```
ChuanqiXu9, line 1114: Please not touch the previous test if the API remains

async_simple/Signal.h
```diff
@@ -0,0 +1,465 @@
...
+    template <typename T = Signal>
+    static std::shared_ptr<T> create() {
+        static_assert(std::is_base_of_v<Signal, T>,
+                      "T should be signal or derived from Signal");
+        return std::make_shared<T>(PrivateConstructTag{});
+    }
+    virtual ~Signal();
+
+private:
+    std::atomic<uint64_t> _state;
```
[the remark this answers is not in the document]
poor-circle: Default state is zero. It means that all signal that has triggered. It was updated at `UpdateState`. If a signal is low 32bits, it can't emit multiple times. I will add some comments.

poor-circle:
@ChuanqiXu9 new conversation is fixed

async_simple/coro/test/LazyTest.cpp
ChuanqiXu9: Can we remain the test?

ChuanqiXu9 (approved):

PR #412, opened 2025-01-16
fix collect heap-use-after-free

poor-circle:
## Why

The prev pull requests #402 may cause heap-use-after-free due to some error operation in collectAny

## What is changing

Fix bug by move variable into await_suspend() function's stack local variable

### DIFF
--- a/async_simple/Signal.h
+++ b/async_simple/Signal.h
@@ -24,11 +24,11 @@
 #include <bit>
 #include <cstdint>
 #include <memory>
+#include <string>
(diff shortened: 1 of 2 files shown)


ChuanqiXu9 (approved):

ChuanqiXu9 (changes requested):
Please fix test failures

poor-circle:
@ChuanqiXu9 all CI passed

ChuanqiXu9 (approved):

PR #414, opened 2025-01-23
fix signal dont triggered bug in collectAll & slot::emplace()

poor-circle:
### `slot::emplace()`

When we call `slot::emplace()`, we check signal state before construct handler. however, when `signal::emit()` called, it will set signal state then set the construct handler.

A possible order:
1. Thread A: `slot::emplace()` check signal state, no triggered now.
2. Thread A: `slot::emplace()` construct signal handler.
3. Thread B: `signal::emit()` set signal state
4. Thread B: `signal::emit()` load signal handler and found it's `nullptr`, emit over
5. Thread A `slot::emplace()` set the signal handler by CAS successful. return true.

Now, `slot::emplace()` will return true and signal handler won't be executed. the signal is loss.

### `collectAny`

`collectAny` should emit signal when the first handler resume, but now it may won't emit it if `downCount` was called in `await_suspend`

### `slot::emplace()`

`slot::emplace()` will construct signal handler before check signal state. 

Because `signal::emit()` will always set state before load signal handler. so if `signal::emit()` load signal handler result is `nullptr`, `slot::emplace()` will always get the signal by check state.

### collectAll and collectAny

Remove `downCount` call in `await_suspend`

### DIFF
--- a/async_simple/coro/Collect.h
+++ b/async_simple/coro/Collect.h
@@ -166,7 +166,7 @@ struct CollectAnyAwaiter {
                 _slot, [c = continuation, e = event, size = input.size()](
                            SignalType type, Signal*) mutable {
                     auto count = e->downCount();
-                    if (count > size + 1) {
+                    if (count == size + 1) {
                         c.resume();
                     }
                 })) {  // has canceled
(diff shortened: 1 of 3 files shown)


poor-circle:
@ChuanqiXu9 All ci passed

async_simple/coro/Collect.h
```diff
@@ -422,11 +425,6 @@ struct CollectAllAwaiter {
             }
             func();
         }
-        _event.setAwaitingCoro(continuation);
```
ChuanqiXu9, line 425: Now the coroutine may be suspended forever if the input was empty.

ChuanqiXu9 (approved):
The variadic version has a static_assert for the inputs size.

PR #456, opened 2026-03-14
ci: fix incorrect clang version for charconv warn

c8ef:
## Why

Based on https://github.com/llvm/llvm-project/commit/381a649fb991eadb0c594de2d8b6166fcc11345a, the warning `character-conversion` was introduced in clang21, not 20, causing the modules CI to fail.


c8ef:
cc @ChuanqiXu9

PR #457, opened 2026-03-15
ci: drop gcc 11 & enable gcc 14

c8ef:
## Why

gcc 10 is the first version to have coroutine support. However, it is still buggy and causing issues in CI.

## What is changing

drop gcc 11 & enable gcc 14


ChuanqiXu9 (approved):
thank you as always

PR #466, opened 2026-03-28
Remove in tree `asio` deps

c8ef:
## Why

Currently, we rely on the in-tree asio, which makes it difficult to upgrade versions.

## What is changing

- Remove the tree in asio. -> 4241e21
- Update bazel builds; update demo-example to accommodate new asio changes. -> f23ac9c
- Update cmake builds. -> e172f66


ChuanqiXu9:
I still like in-tree deps as it is only for examples. And many people have some problems with network, so I don't want to change this. If you want to upgrade its version, I think you can upgrade it too. It might not be so hard.

c8ef:
> I still like in-tree deps as it is only for examples. And many people have some problems with network, so I don't want to change this. If you want to upgrade its version, I think you can upgrade it too. It might not be so hard.

I agree with the network conditions, but we have been using this approach for gtest for two years. I simply copied the approach. 

https://github.com/alibaba/async_simple/blob/ddbd887eb17d30f86772c880b22dbfbb709f395f/cmake/FindGTest.cmake#L10-L15

If users find it difficult to clone the repository, using asio from the package manager would be a better solution. And the newly added cmake file supports this.

ChuanqiXu9 (approved):
Sounds not bad

PR #467, opened 2026-04-04
support uthread on windows with msvc

c8ef:
## Why

related: #451


async_simple/uthread/test/UthreadTest.cpp
```diff
@@ -515,7 +515,7 @@ TEST_F(UthreadTest, testLatchThreadSafe) {
                 auto f = [&taskEx, &taskNotify]() mutable {
                     Latch latch(1u);
                     taskNotify.schedule([latchPtr = &latch]() mutable {
-                        std::this_thread::sleep_for(1us);
```
c8ef, line 518: `std::this_thread::sleep_for` with durations less than one millisecond on Windows gets rounded up to the system timer tick, which is typically around 1ms to 15.6ms, causing tests to run slower on Windows.

async_simple/uthread/internal/thread.cc
```diff
@@ -169,8 +188,13 @@ void thread_context::main() {
...
 #elif defined(__riscv)
     asm(".cfi_undefined ra"); // The return address register in RISC-V is 'ra'
+#elif defined(_MSC_VER)
+    // MSVC uses SEH (structured exception handling) with .pdata/.xdata for
+    // unwinding instead of DWARF CFI directives. The unwind information is
+    // already provided in the MASM assembly files, so no inline annotation is
+    // needed here.
 #else
-#warning "Backtracing from uthreads may be broken"
+#error "Backtracing from uthreads may be broken"
```
ChuanqiXu9, line 197: warning is fine. Uthread is not a necessary part of the library.

c8ef: Unfortunately, `#warning` is standardized in C++23, so it cannot be used in MSVC with C++20 : (
And I believe this line is unreachable for all platforms that support uthread.

ChuanqiXu9: If it is unreachable, let's remove it.

c8ef: Perhaps we could change it to `assert(false, "...")`?

ChuanqiXu9: It is the same thing with #error here. Let's land it, I don't want nitpicking

ChuanqiXu9 (approved):

PR #428, opened 2025-06-10
FutureAwaiter Support Cancellation

poor-circle:
## Why

FutureAwaiter should adapt signal cancellation.

## What is changing

Add support of cancellation for FutureAwaiter.

Now `co_await std::move(future)` could be canceled by signal.

## Example

```cpp
template <typename Callback>
void sum(int a, int b, Callback&& callback) const {
    std::thread([callback = std::move(callback), a, b]() mutable {
        callback(a + b);
    }).detach();
}
async_simple::executors::SimpleExecutor ex1(2);
void test() {
    auto lazy = [&]() -> Lazy<> {
        Promise<int> pr;
        auto fut = pr.getFuture();
        sum(1, 1, [pr = std::move(pr)](int val) mutable {
            std::this_thread::sleep_for(std::chrono::seconds::max());
            pr.setValue(val);
        });
        async_simple::SignalType type = None;
        try {
            co_await std::move(fut);
        } catch (const async_simple::SignalException& e) {
            type = e.value();
        } catch (...) {
        }
        EXPECT_EQ(type, async_simple::Terminate);
    };
    syncAwait(collectAll<async_simple::Terminate>(
        lazy().via(&ex1),
        async_simple::coro::sleep(std::chrono::microseconds{10}).via(&ex1)));
}

```

PR #471, opened 2026-08-03
fix: surface SignalException in FutureAwaiter<void> await_resume

### LINKED ISSUE #470: FutureAwaiter<void> swallows SignalException: co_await Future<void> returns normally when the task is cancelled
## Summary

`co_await Future<void>` does **not** throw `SignalException` when the awaiting task is cancelled — it returns normally instead. `FutureAwaiter<T>` only propagates the cancellation for non-void `T`.

## Root cause

`async_simple/coro/FutureAwaiter.h`, `await_resume()`:

```cpp
auto await_resume() {
    if constexpr (!std::is_same_v<T, void>)
        return std::move(future_.value());
}
```

For `T = void` the function body is **empty**: `future_.value()` is never called.

The cancellation machinery in `await_suspend_cancellable()` relies on `await_resume()` inspecting the future: it replaces `future_` with a ready future carrying `SignalException("FutureAwaiter is cancelled by signal")` and either
- registers a Terminate handler that resumes the coroutine (mid-await cancellation), or
- returns `false` when the signal already fired before the await (pre-fired signal), in which case `await_resume()` runs immediately.

In both paths, for `void` futures the exception is never examined and is silently discarded, so `co_await` completes normally.

## Minimal reproduction

```cpp
#include <async_simple/Promise.h>
#include <async_simple/Signal.h>
#include <async_simple/coro/Collect.h>
#include <async_simple/coro/FutureAwaiter.h>
#include <async_simple/coro/Lazy.h>
#include <async_simple/coro/Sleep.h>
#include <async_simple/coro/SyncAwait.h>
#include <async_simple/executors/SimpleExecutor.h>

using namespace async_simple;

void test() {
    executors::SimpleExecutor ex1(2);
    auto lazy = [&]() -> coro::Lazy<> {
        Promise<void> pr;
        auto fut = pr.getFuture();  // never fulfilled; cancelled externally
        try {
            co_await std::move(fut);
            std::printf("returned normally (BUG)\n");
        } catch (const SignalException& e) {
            std::printf("threw: %d\n", static_cast<int>(e.value()));
        }
        co_return;
    };
    coro::syncAwait(coro::collectAll<SignalType::Terminate>(
        lazy().via(&ex1),
        coro::sleep(std::chrono::microseconds{10}).via(&ex1)));
}
```

Output: `returned normally (BUG)` — the cancellation is silently lost.

## Impact

Any code awaiting a `Future<void>` under `task::TaskManager`-style signal cancellation gets no exception: the coroutine keeps running as if nothing happened (and, e.g., loops forever re-awaiting fresh futures — observed in practice as an infinite spin in a semaphore built on `Promise<void>`). Note that cancellation **does** work correctly for non-void futures (e.g. `Future<int>`, `Future<FetchResult>`), because `value()` is then actually called.

## Suggested fix

Inspect the result for `void` as well:

```cpp
auto await_resume() {
    if constexpr (std::is_same_v<T, void>)
        future_.value();  // surface SignalException even for void
    else
        return std::move(future_.value());
}
```

`future_.value()` on a completed void future is a no-op (empty `Try<void>`), so the non-cancellation path is unaffected.

I'll open a PR with this fix.

deretame:
Fixes #470

## Problem

`co_await Future<void>` does not throw `SignalException` when the awaiting task is cancelled — it returns normally instead. `FutureAwaiter<T>::await_resume()` is:

```cpp
auto await_resume() {
    if constexpr (!std::is_same_v<T, void>)
        return std::move(future_.value());
}
```

For `T = void` the function body is empty, so the cancellation-result future (a ready future carrying `SignalException`, installed by `await_suspend_cancellable()`) is never inspected and the exception is silently discarded — for both the mid-await cancellation path (Terminate handler resumes the coroutine) and the pre-fired-signal path (`tryEmplace` fails, `await_resume()` runs immediately).

## Fix

Inspect the result for `void` as well:

```cpp
auto await_resume() {
    if constexpr (std::is_same_v<T, void>) {
        future_.value();
    } else {
        return std::move(future_.value());
    }
}
```

`future_.value()` on a completed void future is a no-op (empty `Try<void>`), so the non-cancellation path is unaffected; the cancellation path now surfaces `SignalException` as expected.

## Verification

- Minimal reproduction (same shape as PR #428's test but with `Promise<void>` and the promise never fulfilled) now prints `threw: 1` (Terminate) instead of `returned normally (BUG)`.
- Existing `FutureAwaiter` tests still pass.

### DIFF
--- a/async_simple/coro/FutureAwaiter.h
+++ b/async_simple/coro/FutureAwaiter.h
@@ -106,8 +106,15 @@ struct FutureAwaiter {
         return true;
     }
     auto await_resume() {
-        if constexpr (!std::is_same_v<T, void>)
+        if constexpr (std::is_same_v<T, void>) {
+            // Inspect the result even for void: the cancellation path replaces
+            // future_ with a ready future carrying SignalException. Skipping
+            // value() here would silently swallow the cancellation and make
+            // the co_await return normally instead of throwing.
+            future_.value();
+        } else {
             return std::move(future_.value());
+        }
     }
 };
 }  // namespace coro::detail
--- a/async_simple/coro/test/FutureAwaiterTest.cpp
+++ b/async_simple/coro/test/FutureAwaiterTest.cpp
@@ -117,6 +117,51 @@ TEST_F(FutureAwaiterTest, testWithFutureCancel) {
         async_simple::coro::sleep(std::chrono::hours{60}).via(&ex1)));
 }
 
+TEST_F(FutureAwaiterTest, testWithFutureCancelVoid) {
+    async_simple::executors::SimpleExecutor ex1(2);
+    // cancel while awaiting: the Terminate handler resumes the coroutine and
+    // await_resume must surface the SignalException even for Future<void>.
+    auto lazy = [&]() -> Lazy<> {
+        Promise<void> pr;
+        auto fut = pr.getFuture();
+        async_simple::SignalType type = None;
+        try {
+            co_await std::move(fut);
+        } catch (const async_simple::SignalException& e) {
+            type = e.value();
+        } catch (...) {
+        }
+        EXPECT_EQ(type, async_simple::Terminate);
+    };
+    syncAwait(collectAll<async_simple::Terminate>(
+        lazy().via(&ex1),
+        async_simple::coro::sleep(std::chrono::microseconds{10}).via(&ex1)));
+    // pre-fired signal: the slot is already triggered, so the second
+    // co_await's tryEmplace fails and await_resume runs immediately.
+    auto lazy2 = [&]() -> Lazy<> {
+        Promise<void> pr1;
+        auto fut1 = pr1.getFuture();
+        try {
+            co_await std::move(fut1);
+        } catch (const async_simple::SignalException& e) {
+            EXPECT_EQ(e.value(), async_simple::Terminate);
+        }
+        Promise<void> pr2;
+        auto fut2 = pr2.getFuture();
+        async_simple::SignalType type = None;
+        try {
+            co_await std::move(fut2);
+        } catch (const async_simple::SignalException& e) {
+            type = e.value();
+        } catch (...) {
+        }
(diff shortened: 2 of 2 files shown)


ChuanqiXu9 (commented):
LG. But we need a test to merge this.

deretame:
Added a regression test covering both cancellation paths for `Future<void>` (mid-await Terminate handler + pre-fired signal with `tryEmplace` failure), verified locally with MSVC: fails on the unfixed code, passes with the fix.

Regarding the failing CI checks — these look like pre-existing infrastructure issues unrelated to this change (which only touches `async_simple/coro/FutureAwaiter.h`):

1. **Windows CMake (Release, x86)**: fails at configure time — the workflow hardcodes `-G "Visual Studio 17 2022"`, but the new `windows-2025-vs2026` runner image (2026-07-14) only ships VS 2026, so CMake reports "Generator ... could not find any instance of Visual Studio". The code is never compiled.
2. **Windows Bazel**: `benchmarks/Lazy.bench.cpp(26)` fails with MSVC `C2999: maximum template instantiation depth of 1000 exceeded` — the N=1000 `lazy_fn` chain benchmark hits MSVC's default `/templateDepth` limit under the new MSVC 14.51 toolchain. That file is untouched by this PR.
3. **macOS Bazel**: fails inside googletest's own header (`gtest-printers.h:528`) with `-Werror,-Wcharacter-conversion` on a `char8_t` -> `char32_t` conversion — a googletest/clang version mismatch. Note the CMake path already works around this (`FindGTest.cmake` adds `-Wno-error=character-conversion` for clang 21+), while the Bazel path uses a different googletest from BCR.

The same Windows jobs were green on PR #468 (April), so this regressed with the runner image update, not with this PR. All Linux/macOS CMake jobs pass, and the test suite runs clean locally on MSVC 14.51.

ChuanqiXu9:
Got it. Thank you very much!

PR #476, opened 2026-08-26
Corrected the docs description of alignment requirement for Lazy<T>

csmyx:
## Why
According to this static assert, the alignment requirement for type 'T' in 'Lazy<T>' should **not** exceed 'alignof(std::max_align_t)'.
``` cpp
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wgnu-alignof-expression"
    static_assert(alignof(T) <= alignof(std::max_align_t),
                  "async_simple doesn't allow Lazy with over aligned object");
#endif
```

## What is changing

Corrected the description of the alignment requirement for type 'T' in 'Lazy<T>' that it can **not** exceed 'alignof(std::max_align_t)'.

### DIFF
--- a/docs/docs.en/Lazy.md
+++ b/docs/docs.en/Lazy.md
@@ -26,7 +26,7 @@ Lazy<int> task2(int x) {
 ## Alignment Requirement
 
 Due the limitation of ABI, Compiler Implementation and the usage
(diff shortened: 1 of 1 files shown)


ChuanqiXu9 (approved):
lgtm, thanks

PR #477, opened 2026-09-11
Guard benchmark option and add regression coverage

ajavanma:
This PR makes benchmark discovery optional in CMake and adds regression coverage.

- Add ASYNC_SIMPLE_ENABLE_BENCHMARKS option default ON in root CMakeLists.
- Only find and add benchmarks/ subdir when option is enabled.
- Document behavior and usage in README.
- Add regression test ensuring --clean-ctest with BENCHMARK option disabled does not touch benchmark discovery.


CMakeLists.txt
```diff
@@ -39,11 +39,14 @@ find_package(Threads REQUIRED)
 target_link_libraries(async_simple_header_only INTERFACE Threads::Threads)
 find_package(Aio QUIET)
 
-find_package(Benchmark)
-
 option(ASYNC_SIMPLE_ENABLE_TESTS "Build the tests" ON)
+option(ASYNC_SIMPLE_ENABLE_BENCHMARKS "Build benchmarks when Benchmark is available" ON)
```
qicosmos: Default OFF is better.

